PR #3, opened 2016-05-08
Pinning messages docs updated

blubbi321:
With small note about the default for message pinning settings. Might confuse non-admin users otherwise as everybody is allowed to pin in Slack.

### DIFF
--- a/6. User Guides/Pinning Messages/README.md
+++ b/6. User Guides/Pinning Messages/README.md
@@ -1,3 +1,5 @@
 # Pinning Messages
 
-Pin the most important messages of the channel
+Pin the most important messages of the channel for everybody to see.
+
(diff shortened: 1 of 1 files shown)


frdmn:
That's a useful hint. Thanks 👍 

Waiting for another team member to confirm.

geekgonecrazy:
LGTM 👍

sampaiodiego:
oops, too late. :disappointed: 

so, actually the `pin-message` permission is given by default to `admin`, `owner` and `moderator` roles..

so, this sentence is not correct:

> only enabled for admin users by default

frdmn:
@sampaiodiego Fixed in a0e2d642b16d564222e23ede3da9d40d7bb6c427:

```
Note that this is only enabled for admins, owners and moderators by default [...]  
```

PR #7, opened 2016-05-13
Windows Guide Update

killianhale:
Added a list of the windows builds and referenced them inside the guide instead of the official releases page


geekgonecrazy:
LGTM

PR #23, opened 2016-06-07
GitHub.md: Input script, switch out image URLs for inline SVGs

xenithorb:
Please don't merge yet. My intention was originally to keep the embedded SVG in <svg> format, but I couldn't figure out how to get `Buffer` working with meteor. I kept getting `ReferenceError: Buffer is not defined` - which was when I realized I was following NodeJS instructions :/ 

If someone can figure out the proper way to encode the svg data into base64 on the fly, let me know. 

My original assumtion was that i could use:

```
var svg_item_1 = Buffer('<svg></svg>', 'ascii'); 
...
    author_icon: svg_b64_pre + svg_item_1.toString('base64');
```

but like I said, that didn't work? 

---

The original intention here was to include the `commit` icon because previously only the pull request icon was being used. Issue icon will be coming next.

### DIFF
--- a/4. Administrator Guides/Integrations/GitHub.md
+++ b/4. Administrator Guides/Integrations/GitHub.md
@@ -11,6 +11,13 @@ We can do 2 types of integrations with GitHub:
 ```javascript
 /* exported Script */
 
+// Begin embedded images
+const gh_cmit_svg = '<svg xmlns="http://www.w3.org/2000/svg" aria-hidden="true" class="octicon octicon-git-commit" height="16" version="1.1" viewBox="0 0 14 16" width="14"><path d="M10.86 7c-.45-1.72-2-3-3.86-3-1.86 0-3.41 1.28-3.86 3H0v2h3.14c.45 1.72 2 3 3.86 3 1.86 0 3.41-1.28 3.86-3H14V7h-3.14zM7 10.2c-1.22 0-2.2-.98-2.2-2.2 0-1.22.98-2.2 2.2-2.2 1.22 0 2.2.98 2.2 2.2 0 1.22-.98 2.2-2.2 2.2z"></path></svg>';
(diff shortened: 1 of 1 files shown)


frdmn:
Since SVGs is just XML and no binary, why not just store it as a string instead?

```
var svg_item_1 = '<svg xmlns="http://www.w3.org/2000/svg" aria-hidden="true" class="octicon octicon-git-commit" height="16" version="1.1" viewBox="0 0 14 16" width="14"><path d="M10.86 7c-.45-1.72-2-3-3.86-3-1.86 0-3.41 1.28-3.86 3H0v2h3.14c.45 1.72 2 3 3.86 3 1.86 0 3.41-1.28 3.86-3H14V7h-3.14zM7 10.2c-1.22 0-2.2-.98-2.2-2.2 0-1.22.98-2.2 2.2-2.2 1.22 0 2.2.98 2.2 2.2 0 1.22-.98 2.2-2.2 2.2z"></path></svg>';
```

xenithorb:
Because my options not knowing the code in-depth were to pass it to `author_icon:` 

The result if it's not done this way is that you end up with:

```
<img src="<svg aria-hidden="true" class="octicon octicon-issue-opened" height="16" version="1.1" viewBox="0 0 14 16" width="14"><path d="M7 2.3c3.14 0 5.7 2.56 5.7 5.7s-2.56 5.7-5.7 5.7A5.71 5.71 0 0 1 1.3 8c0-3.14 2.56-5.7 5.7-5.7zM7 1C3.14 1 0 4.14 0 8s3.14 7 7 7 7-3.14 7-7-3.14-7-7-7zm1 3H6v5h2V4zm0 6H6v2h2v-2z"></path></svg>">
```

Which doesn't work, Instead of:

```
<img src="data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIGFyaWEtaGlkZGVuPSJ0cnVlIiBjbGFzcz0ib2N0aWNvbiBvY3RpY29uLWdpdC1jb21taXQiIGhlaWdodD0iMTYiIHZlcnNpb249IjEuMSIgdmlld0JveD0iMCAwIDE0IDE2IiB3aWR0aD0iMTQiPjxwYXRoIGQ9Ik0xMC44NiA3Yy0uNDUtMS43Mi0yLTMtMy44Ni0zLTEuODYgMC0zLjQxIDEuMjgtMy44NiAzSDB2MmgzLjE0Yy40NSAxLjcyIDIgMyAzLjg2IDMgMS44NiAwIDMuNDEtMS4yOCAzLjg2LTNIMTRWN2gtMy4xNHpNNyAxMC4yYy0xLjIyIDAtMi4yLS45OC0yLjItMi4yIDAtMS4yMi45OC0yLjIgMi4yLTIuMiAxLjIyIDAgMi4yLjk4IDIuMiAyLjIgMCAxLjIyLS45OCAyLjItMi4yIDIuMnoiPjwvcGF0aD48L3N2Zz4=">
```

Which does..

frdmn:
Ah, I didnt think that far that this ends up in a concated `` src.

I can't think of any other more elegant solution for this then :/

xenithorb:
The main problem is that I don't know of a meteor-compatible way to base64 encode a string. The recommended way for NodeJS was to use `Buffer().toString('base64')` - So I backed out of encoding on the fly and decided to just store the base64 as a proof-of-concept.

xenithorb:
Ah, as it turns out, it looks like you might be able to pass it as `utf8` and then insert the SVG XML: https://css-tricks.com/probably-dont-base64-svg/

You were onto something after all...

xenithorb:
Proof that it works:

screenshot_20160607_080443

xenithorb:
I've been using this all day and haven't seen any issues with it. I think it's ready for prime-time if you'll have it. I prefer it this way so that it's not making calls to external sites for the images. @RocketChatApp

frdmn:
> I prefer it this way so that it's not making calls to external sites for the images. 

Totally agree. LGTM :)

PR #32, opened 2016-06-18
Added the env Accounts_UseDNSDomainCheck=false to docker example

TheReal1604:
Added Description for the env "Accounts_UseDNSDomainCheck=false" that is need behind a proxy in order to create the first admin account.


Sing-Li:
Thanks, @TheReal1604 .    

This will _really_ help!    We get **so** many support cases with this problem.

geekgonecrazy:
LGTM 👍

PR #40, opened 2016-07-01
Doc to explain regex searching

RickCogley:
Started a doc to try to explain regex in a simple way, to help people get started, and, to let CJK users know that they can search CJK strings this way too. 

From this gist: https://gist.github.com/RickCogley/e70679653771c0ae898d41a24e29723a

Signed-off-by: Rick Cogley rick.cogley@esolia.co.jp


Sing-Li:
LGTM

PR #46, opened 2016-07-11
Adding tag v0.1.4 on hubot container

nicolaspeixoto:
Latest tag was built using version v1.x.x of adapter and it is experimental, and in anticipation of a performance oriented sweeping change within the core - which has not happened yet.


geekgonecrazy:
@nicolaspeixoto thanks for the PR.  Yeah timing didn't work out like expected.

PR #47, opened 2016-07-12
Oplog Tailing

codyaray:
This enables Oplog Tailing by default for the high-performance example


Sing-Li:
May need to add instructions on how to setup mongo for oplog tailing.   By default,  I don't think it is enabled.

codyaray:
You mean creating a replica set? I think that's the only thing you have to do.

PR #49, opened 2016-07-16
Update mongodb-org package for Debian to 3.2

unixfox:
I confirm that Rocket chat works on MongoDB 3.2.

### DIFF
--- a/3. Installation/4. Manual Installation/Debian/README.md
+++ b/3. Installation/4. Manual Installation/Debian/README.md
@@ -11,7 +11,7 @@ This guide explains how to deploy your own Rocket.Chat instance to a Debian Whee
 
 ```shell
 # SYSTEM CONFIGURATION
-sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 7F0CEB10
+sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv EA312927
 # Here use wheezy even if you are installing it on Jessie. 
 # since Jessie repository of Mongodb does NOT includes mongodb 3.2
 echo "deb http://repo.mongodb.org/apt/debian wheezy/mongodb-org/3.2 main" | sudo tee /etc/apt/sources.list.d/mongodb-org-3.2.list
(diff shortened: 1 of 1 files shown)


frdmn:
LGTM!

frdmn:
Just successfully tested the installaiton of RocketChat on a DigitalOcean Debian 8.5 droplet. 

One little thing, as I tried to initialize the Mongo replication (`rs.initiate()`), I did run into the following error:

```
{
    "info2" : "no configuration explicitly specified -- making one",
    "me" : "0.0.0.0:27017",
    "errmsg" : "couldn't initiate : can't find self in the replset config",
    "ok" : 0
}
```

Workaround for this, is to comment out the `bindip` option in `/etc/mongod.conf`:

```
net:
  port: 27017
#  bindIp: 127.0.0.1
```

@unixfox Can you please check if you had to do this as well?

frdmn:
@Sing-Li If I remember correctly you said something yesterday about 3.2 doesn't work yet. What exactly doesn't work? Can't find any issues right now.

unixfox:
@frdmn I don't have any problem with the default configuration (I just added smallfiles):

```
# mongod.conf
# for documentation of all options, see:
#   http://docs.mongodb.org/manual/reference/configuration-options/
# Where and how to store data.
storage:
  dbPath: /var/lib/mongodb
  journal:
    enabled: true
#  engine:
  mmapv1:
    smallFiles: true
#  wiredTiger:
# where to write logging data.
systemLog:
  destination: file
  logAppend: true
  path: /var/log/mongodb/mongod.log
# network interfaces
net:
  port: 27017
  bindIp: 127.0.0.1
#processManagement:
#security:
#operationProfiling:
#replication:
#sharding:
## Enterprise-Only Options:
#auditLog:
#snmp:
```

Sing-Li:
We are based on Meteor 1.2.1 -  and Meteor 1.2.1 has not been extensively tested on Mongo 3.2 -  so the potential problems are documented on these two tickets.

https://github.com/meteor/meteor/issues/5809

https://github.com/meteor/meteor/issues/6957

I'm reluctant to recommend 3.2 usage at this time.  We should be upgrading to Meteor 1.4 in a little while - at that time we will "catch up" with the latest mongo.

We can merge this down after our upgrade to Meteor 1.4, thanks.

PR #50, opened 2016-07-17
Alternative script and customization instructions for Github integration

alexweissman:
I prefer an integration that looks more like the autolinker messages.  So, I created a custom script that generates messages of this form:

github-1

Note that the message itself is posted by a bot, but the actual Github user's avatar is contained within the message as a thumbnail.


Sing-Li:
LGTM

I'm not 100% sure the PNG will be picked up by our doc system where they are.

But thanks @alexweissman  !

alexweissman:
Cool, I was just following the same format as I saw in https://github.com/RocketChat/Rocket.Chat.Docs/tree/master/4.%20Administrator%20Guides/Livechat.

Sing-Li:
Alright!

PR #53, opened 2016-07-23
README.md - Landing page links to chapters

sunnyolives:
Optimally when you click on docs - you immediately know where to look to find your answer. I think adding these links to the top would improve the landing page, because you don't have to scroll to find the user guide, for instance. Ideally the screenshot would be to the right of the link, or the links would be buttons spanning horisontally, because in my proposal I think the page looks imbalanced with narrow lists on top and the huge image below (this is on a wide screen device). As the page looks before my proposal, I think it looks better on mobile, the image is smaller.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,12 @@
 # Rocket.Chat Docs
+## What would you like to do?
+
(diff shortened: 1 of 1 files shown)


engelgabriel:
Thanks @sunnyolives

PR #54, opened 2016-07-24
update email config documentation

alexweissman:
- Set SMTP settings directly in Admin interface
- Reboot chat server


engelgabriel:
Thanks @alexweissman

PR #92, opened 2016-09-21
Update manual installation on Ubuntu/Debian

baierjan:
Added replicaSet option to avoid standalone connection (see https://docs.mongodb.com/manual/reference/connection-string/#replica-set-option)


geekgonecrazy:
LGTM  Thanks for the contribution!

geekgonecrazy (approved):

PR #117, opened 2016-11-08
Update README.md

harishreddy-m:
Empty file is not good


geekgonecrazy:
Please open an issue here for that.  I agree empty isn't preferable.  :)

geekgonecrazy (approved):

geekgonecrazy:
I changed it so its a more positive inviting message.  But like I said please open an issue for missing documentation.

frdmn (commented):
👍

PR #120, opened 2016-11-11
Improve Caddy SSL reverse proxy

j8r:
A little improvement concerning Caddy and reverse proxying.
I've also added the good `Caddyfile` sample stolen from @geekgonecrazy


geekgonecrazy:
@j8r awesome!  One concern. Added below.

geekgonecrazy (changes requested):
We need to align any links consuming the "Configuring SSL Reverse Proxy with Nginx" link and update to new URL

geekgonecrazy (approved):

PR #126, opened 2016-11-21
Remove StartSSL instructions

frdmn:
As @eionrobb mentioned in #general, [StartCom (StartSSL) got removed](https://security.googleblog.com/2016/10/distrusting-wosign-and-startcom.html) so we probably should remove the instructions in the docs as well.


EionRobb:
Is it worth pointing to Let's Encrypt instead?

frdmn:
Not a bad idea, however I have no experience at all with LetsEncrypt on Windows. 

Well, something I will try out next weekend. :)

geekgonecrazy (approved):

geekgonecrazy:
I'll go ahead and merge this so we get the bad instructions out.  👍

PR #130, opened 2016-11-25
Add info about console.log

milep:
May need a better phrasing.

### DIFF
--- a/5. Administrator Guides/Integrations/README.md
+++ b/5. Administrator Guides/Integrations/README.md
@@ -20,7 +20,7 @@ The class should have a method called **process_incoming_request**, this method 
 
 The **process_incoming_request** method should return an object with a property **content** containing a valid Rocket.Chat message or an object with a property **error** that will be returned as the response of the request in JSON format and status code **400**.
 
-You can use the **console** to log and view the logs from Administration -> View Logs.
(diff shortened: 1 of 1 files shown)


5. Administrator Guides/Integrations/README.md
```diff
@@ -20,6 +20,8 @@ The class should have a method called **process_incoming_request**, this method
 
 The **process_incoming_request** method should return an object with a property **content** containing a valid Rocket.Chat message or an object with a property **error** that will be returned as the response of the request in JSON format and status code **400**.
 
+You can use the **console** to log and view the logs from Administration -> View Logs.
```
geekgonecrazy: Maybe we should include a link to: https://developer.mozilla.org/en-US/docs/Web/API/Console/log

geekgonecrazy (changes requested):

geekgonecrazy:
Thanks for contributing!  Left a note.  Might be useful to give some context as to what console is for those that don't understand :)

[edit] Made the adjustments.

5. Administrator Guides/Integrations/README.md
engelgabriel: @geekgonecrazy can you add your suggestion directly?

geekgonecrazy: @engelgabriel Ah looks like that is enabled now!  Sweet.  Changes made.

geekgonecrazy (dismissed):

engelgabriel (approved):

PR #134, opened 2016-12-01
Updated template Parameter to MONGODB_PASSWORD

phil-pona:
The current Template parameter in the openshift template is MONGODB_PASSWORD not MONGODB_PASS


graywolf336 (changes requested):
What about changing the parameters from being italicized to being code highlighted?

graywolf336 (approved):
Awesome, thanks for the pull request and making the changes!

PR #136, opened 2016-12-01
Clarified methods for HTTP requests POST/GET

SloMusti:
Improved the documentation, such that it is clear which function requires what method and a bit more about the operation of the API.


geekgonecrazy (approved):

geekgonecrazy:
@SloMusti thanks for contributing!

PR #141, opened 2016-12-08
Fixed: missed the date object in the parameter

rafaelks:


graywolf336 (approved):
Ah I should have caught that, good catch. 👍

PR #142, opened 2016-12-09
Rest api additions

graywolf336:
This should be merged the same time https://github.com/RocketChat/Rocket.Chat/pull/5140 gets merged.


sampaiodiego (approved):
looks awesome.. great work :clap: 

As a furter suggestion I think we should add a `Permissions required` column on each endpoint page.

graywolf336:
@sampaiodiego I agree, the permissions should be added. Would be a nice contribution for someone to add who wants to get familiar with the source of Rocket.Chat

PR #155, opened 2016-12-19
add Less Code Styleguide

karlprieb:


sampaiodiego:
@karlprieb you don't need to change the file names anymore to change their order..

just add on top of the files you want to (re)order:
```
---
order: 99
---
```

engelgabriel (approved):

PR #175, opened 2017-01-12
Add note to Uncomment console.log to debug in Telegram Integration

robbyoconnor:


geekgonecrazy:
@robbyoconnor I might be missing something.  But why do we need the console.log uncommented?  I would think this would only be needed in a development environment

robbyoconnor:
@geekgonecrazy either it needs to be commented, or the comment above it needs to change :)

robbyoconnor:
@geekgonecrazy fixed it. Prior it looked like an oversight -- I copied/pasted straight from The integration I was using. Now it's clearer. The console.log is quite useful.

geekgonecrazy (approved):
Thanks :+1: I think that's totally acceptable.  Just didn't want loose console.log's as a default :grin:

robbyoconnor:
Well it's there -- they just have to read it..

PR #179, opened 2017-01-15
Update systemd services

thiesschneider:
Added instructions for systemd

### DIFF
--- a/3. Installation/3. Docker Containers/systemd/README.md
+++ b/3. Installation/3. Docker Containers/systemd/README.md
@@ -8,19 +8,17 @@ You need to have [docker](https://docs.docker.com/linux/started/) installed.
 
 ## How to run Rocket.Chat on systemd.
 
-Create the docker-container for the mongo-init-replica. It will remain in an exited state. Do NOT remove this!
+1. Create two service files: mongo.service and rocketchat.service
+2. Enable the service files
+3. Run the mongo.service
+4. Create the mongo-init-replica to enable opolog
+5. Run the rocketchat.service
 
-```
-docker run \
-      --name mongo-init-replica \
-      --link mongo:mongo \
-      --net=rocketchat_default \
-      mongo:3.2 \
-      mongo mongo/rocketchat --eval "rs.initiate({ _id: 'rs0', members: [ { _id: 0, host: 'localhost:27017' } ]})"
-```
-Then create two systemd files. One for the database and one for rocketchat.
+If you reboot the server, the services will start automatically in the right order.
 
-If you restart the services come up automatically and in the right order.
+### The docker-container for the mongo-init-replica will remain in an exited state. Do NOT remove the container `mongo-init-replica`!
+
+## Service Files
 
 mongo.service:
 ```
@@ -50,7 +48,8 @@ ExecStart=/usr/bin/docker run \
 ExecStop=-/usr/bin/docker kill mongo
 ExecStop=-/usr/bin/docker rm mongo
 ```
-and the rocketchat.service:
+
+rocketchat.service:
 ```
 [Unit]
 Description=rocketchat
@@ -82,6 +81,17 @@ ExecStart=/usr/bin/docker run \
 ExecStop=-/usr/bin/docker kill rocketchat
 ExecStop=-/usr/bin/docker rm rocketchat
 ```
+
+command to build the mongo-init-replica container (only run once):
+```
+docker run \
+      --name mongo-init-replica \
+      --link mongo:mongo \
+      --net=rocketchat_default \
+      mongo:3.2 \
+      mongo mongo/rocketchat --eval "rs.initiate({ _id: 'rs0', members: [ { _id: 0, host: 'localhost:27017' } ]})"
+```
+
 
 ## Reverse Proxy
 
--- a/3. Installation/3. Docker Containers/systemd/README.md
+++ b/3. Installation/3. Docker Containers/systemd/README.md
@@ -4,20 +4,101 @@ Deploy [Rocket.Chat](https://github.com/RocketChat/Rocket.Chat) to Linux that ru
 
 ## Prerequisite
 
-You need to have [docker](https://docs.docker.com/linux/started/) and [docker-compose](http://docs.docker.com/compose/) installed.
+You need to have [docker](https://docs.docker.com/linux/started/) installed.
 
 ## How to run Rocket.Chat on systemd.
 
+Create three systemd files. One for the database, one for the database replica and one for rocketchat.
+
+mongo.service:
 ```
-git clone https://github.com/RocketChat/DockerFiles.git
-cd systemd
-cp ./*@* /etc/systemd/system && systemctl daemon-reload
-mkdir -p /data/domains/
-cd ..
-cp GenericLinux /data/domains/example.org
-vi /data/domains/example.org/docker-compose.yml # modify `ROOT_URL` and `MAIL_URL`
-systemctl enable universal@example.org
-systemctl start universal@example.org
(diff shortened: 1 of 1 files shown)


3. Installation/3. Docker Containers/systemd/README.md
```diff
@@ -4,20 +4,101 @@ Deploy [Rocket.Chat](https://github.com/RocketChat/Rocket.Chat) to Linux that ru
...
+Description=mongo
+Requires=docker.service
+After=docker.service
+
+[Service]
+EnvironmentFile=/etc/environment
+User=dockeruser
+Restart=always
+TimeoutStartSec=0
+ExecStartPre=-/usr/bin/docker kill mongo
```
geekgonecrazy, line 34: Have you been using this in production?  This kind of makes me nervous.  Actually blowing away the mongo container every time.  I know below you have volume mounts... But seems like you would want to make sure your image stayed exactly the same.

3. Installation/3. Docker Containers/systemd/README.md
```diff
@@ -4,20 +4,101 @@ Deploy [Rocket.Chat](https://github.com/RocketChat/Rocket.Chat) to Linux that ru
...
-cd systemd
-cp ./*@* /etc/systemd/system && systemctl daemon-reload
-mkdir -p /data/domains/
-cd ..
-cp GenericLinux /data/domains/example.org
-vi /data/domains/example.org/docker-compose.yml # modify `ROOT_URL` and `MAIL_URL`
-systemctl enable universal@example.org
-systemctl start universal@example.org
+[Unit]
+Description=mongo-init-replica
```
geekgonecrazy: So this one would run every time?  I'm wondering if there could be adverse effects by running this everytime it starts up?

3. Installation/3. Docker Containers/systemd/README.md
```diff
@@ -4,20 +4,101 @@ Deploy [Rocket.Chat](https://github.com/RocketChat/Rocket.Chat) to Linux that ru
...
+ExecStartPre=-/usr/bin/docker kill mongo-init-replica
+ExecStartPre=-/usr/bin/docker rm mongo-init-replica
+ExecStartPre=-/usr/bin/docker pull mongo:3.2
+
+ExecStart=/usr/bin/docker run \
+      --name mongo-init-replica \
+      --link mongo:mongo \
+      --net=rocketchat_default \
+      mongo:3.2 \
+      mongo mongo/rocketchat --eval "rs.initiate({ _id: 'rs0', members: [ { _id: 0, host: 'localhost:27017' } ]})"
```
geekgonecrazy: I know this adds complications to this... But, maybe there is something that could be ran to check if this has been done already?

geekgonecrazy (commented):
Thanks for this!  Very much needed.  Just a few comments / concerns inline.

3. Installation/3. Docker Containers/systemd/README.md
thiesschneider: I am using this in production and it is killing and removing this everytime it is recreated. The persistent data is in the volume. I also update the mongo image if something is pushed (i think it is the same way for docker-compose if i pull, and make up -d)

3. Installation/3. Docker Containers/systemd/README.md
thiesschneider: You do not need to run it everytime. Thats right. Perhaps it is better to make this a one time command in the beginning.

3. Installation/3. Docker Containers/systemd/README.md
thiesschneider: In my setup it is starting everytime and nothing is breaking :)
And I don't know how to check it ^^ 
So I need a little help here ;)

graywolf336:
@geekgonecrazy What's the status on your review?

geekgonecrazy:
@graywolf336 Only the rs.initiate() really concerns me.  The fact that it can run more then once.  We need to make sure this won't be harmful in anyway.  Ideally this would run once, and should never run again.

I think we should remove that as a service.  I think we should guide them like this:
1. Add mongo service.
2. Enable / Start Service
3. Rs.initiate() manually not via a service
4. Add Rocket.Chat service
5. Enable / Start Service

thiesschneider:
yep I think this would be the right way.
I improve this part tomorrow.

thiesschneider:
@geekgonecrazy @graywolf336 What's the status on your reviews?

geekgonecrazy:
@thiesschneider looks better.  We just need instructions to say to add both service files first.  Enable both.  Start mongodb.  Run that command and then start Rocket.Chat. 

Let me know if you need help.  I can adjust if need be

thiesschneider:
@geekgonecrazy I hope its now updated correctly :)

geekgonecrazy:
Thanks for your patience and adjustments!  I know annoying it can be :)

3. Installation/3. Docker Containers/systemd/README.md
```diff
@@ -4,22 +4,103 @@ Deploy [Rocket.Chat](https://github.com/RocketChat/Rocket.Chat) to Linux that ru
...
+1. Create a docker network `docker network create rocketchat_default`
+2. Create two service files: mongo.service and rocketchat.service
+3. Enable the service files
+4. Run the mongo.service
+5. Create the mongo-init-replica to enable opolog
+6. Run the rocketchat.service
+
+If you reboot the server, the services will start automatically in the right order.
+
+### The docker-container for the mongo-init-replica will remain in an exited state. Do NOT remove the container `mongo-init-replica`!
```
geekgonecrazy: I'm probably over looking something so I apologise if I am.  But why do they need to leave this container around?

thiesschneider: @geekgonecrazy no problem!

I might be wrong, but does the mongo database not save data in this container like in an data container?

geekgonecrazy: @thiesschneider since you start the container and connect to the other mongo container, no data will be saved in this containers volume.  So completely disposable, you are only linking for the purposes of allowing the containers to talk, not inheriting it's data volumes.  

Also since you are actually mounting the mongo container to a local disk directory you are even doubly safe.

thiesschneider: @geekgonecrazy I did not delete my container because of some instruction like this: http://www.sohamkamani.com/blog/2016/06/30/docker-mongo-replica-set/

> In case we remove one of our containers by mistake, the data would also vanish. Using Docker volumes and setting the appropriate --dbpath when running mongod would prevent this from happening.

I am running this on CoreOS so I also safe my Volumes in Data Containers which are also in the exited state and should not be removed. So perhaps someone else knows the answer or I will try this in a test environment...

thiesschneider:
@geekgonecrazy any news on this?

3. Installation/3. Docker Containers/systemd/README.md
geekgonecrazy: If you could, lets go ahead and remove this line.  I've tested a couple of times on my machine.

The command starts a new container with mongo inside of it.  But since you are running the mongo command directly, the built in mongod in that container does not start.  Instead mongo connects to the other container(your running mongo container), executes the command.  Nothing touches the files / disk on the temporary container at all, so it is completely safe to delete.

thiesschneider:
@geekgonecrazy thx for testing. Done :)

geekgonecrazy (approved):

geekgonecrazy:
@thiesschneider thanks again!

PR #187, opened 2017-01-18
Remove invalid method that no longer exists

graywolf336:
Closes: https://github.com/RocketChat/Rocket.Chat/issues/5633


sampaiodiego:
the `rooms/get` method still exists. as you can see [here](https://github.com/RocketChat/Rocket.Chat/blob/develop/server/publications/room.js#L39)

graywolf336:
:laughing: I was just going off the issue

PR #199, opened 2017-01-27
Added Minio as top level storage option

nitisht:


geekgonecrazy:
@NitishT yeah much better suited as a top level storage solution :smile:

I'm very impressed with minio!  Thanks for the work you and the rest of your team have put into making this great solution! :+1:

geekgonecrazy (approved):
No real reason for it not to be its own top level.  Especially when it is its own product not a sub of Amazon or Amazon s3

nitisht:
Thanks for the kind words @geekgonecrazy

PR #203, opened 2017-01-31
Update AutoSSL.md

kurobeats:


geekgonecrazy:
@kurobeats Both methods should be valid.  Curious why you feel this way is better?

kurobeats:
@geekgonecrazy `sudo rocketchat-server.initcaddy` did work on Ubuntu but not on Debian whereas `sudo snap run rocketchat-server.initcaddy` worked on both.

geekgonecrazy (approved):

geekgonecrazy:
Changed it to make it an alternative.  As this is under the Ubuntu section.  So this is an Ubuntu doc.

PR #204, opened 2017-02-02
Use the upstream definition for proxy pass

sebastianrakel:
You didn't have to provide an IP at proxy_pass because a upstream is defined


geekgonecrazy (approved):
yes definitely should be using the backend defined

PR #228, opened 2017-02-21
MongoDB install needs to occur as administrator

catmanjan:
Instructions also included a cd to a folder that does not necessarily exist.


geekgonecrazy (approved):

geekgonecrazy:
@catmanjan thanks for this change :+1:

PR #236, opened 2017-03-03
In history API docs, include groups.history endpoint

ragesoss:


graywolf336:
Group history is already defined though: https://rocket.chat/docs/developer-guides/rest-api/groups/history#group-history

ragesoss:
Whoops! I keep on getting confused over and over again about the split endpoints for groups vs. channels.

PR #239, opened 2017-03-04
Open up the possibility for more innovation on OTR

Sing-Li:
per recommendations from john.sinteur's radically open security group as well as mitar

### DIFF
--- a/1. Contributing/Google Summer of Code/README.md
+++ b/1. Contributing/Google Summer of Code/README.md
@@ -16,9 +16,9 @@ For general information, and some real-time help on next steps, please visit our
 * **Goal/Deliverable:** The foundation for all Rocket.Chat users/bots/services to optionally communicate with one another, across all connected instances.
 
 
-### Improved Off The Record Messaging (OTR)
+### Improved End-to-End Encryption Implemenation (currently feature-named OTR)
 * **Mentors:** Marcelo Schmidt ([@marcelo.schmidt](https://github.com/marceloschmidt))  Mitar Milutinovic ([@mitar](https://github.com/mitar))
-* **Description:** Taking and improving our OTR implementation. Perhaps replace with more modern protocols, such as Signal. 
+* **Description:** Taking and improving our end-to-end encryption implementation. Perhaps replace with more modern protocols, such as Signal. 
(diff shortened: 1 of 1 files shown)


mitar:
I think we should rename OTR to end-to-end encryption. OTR is just a specific way to implement it (with forward secrecy), and also an existing name of the protocol (which we are currently not even following, but have our own). What we want is *an* end-to-end encryption. And then we can later on decide which exactly form to use. I also agree that Signal protocol is probably the best way to go for end-to-end encryption because it allows both forward secrecy mode (where both parties have to be online), and not, group encryption, and so on.

Sing-Li:
You, the expert.  Updating now :+1:

1. Contributing/Google Summer of Code/README.md
```diff
@@ -16,11 +16,11 @@ For general information, and some real-time help on next steps, please visit our
...
 
-### Improved Off The Record Messaging (OTR)
+### Improved End-to-End Encryption Implemenation (currently feature-named OTR)
 * **Mentors:** Marcelo Schmidt ([@marcelo.schmidt](https://github.com/marceloschmidt))  Mitar Milutinovic ([@mitar](https://github.com/mitar))
-* **Description:** Taking and improving our OTR implementation
-* **Desirable Skills:** Some javascript coding skills and familiarity with cryptography.
-* **Goal/Deliverable:** The ability to have secure OTR conversations
+* **Description:** Taking and improving our end-to-end encryption implementation. Perhaps replace with more modern protocols, such as Signal. 
+* **Desirable Skills:** Some Javascript coding skills and familiarity with cryptography.
+* **Goal/Deliverable:** The ability to have secure conversations
```
mitar: So this is a question how open we want this for students. One option would be to simply say, we want signal protocol. If we do not specify it so precisely, then we should probably be more precise here and say something out of many possibilities like  that we want secure conversations between two parties, in group chats, when all parties are online, when some are offline, secure file transfers, and so on.

Sing-Li: @mitar -  Sure. I think we want to be as flexible and open as possible (and still keep the description brief and concise).  Can you please modify it in place to cater for the latter?

engelgabriel (approved):

PR #252, opened 2017-03-14
User guide changes add Jira outbound integration

Rohlik:
I think that right category is "Integration" not "Import".


Rohlik:
Also adding info about Jira integration.
PS: Sorry for spam.

Rohlik:
Still no review??

geekgonecrazy:
@marceloschmidt as most familiar with slackbridge, can you take a look?

geekgonecrazy:
Actually disregard.  The bulk of this is Jira which looks fine.  

The slackbridge re-categorization idk if I agree with.  It's currently not in Rocket.Chat treated as an integration.  Its more of an import / export as it's importing messages into rocket.chat

geekgonecrazy (changes requested):
Other then the re-categorizing of slack bridge all the other changes look good :+1:  Please revert that for now. 

The slackbridge is intended as a shim to allow groups to export slack import into Rocket.Chat and then run the bridge to keep the import going until they are able to get all of their users moved over to Rocket.Chat.

While in a way its an integration.  It purpose really is to import.

If anything at all... I'd say we need a bridge section when we get other bridges in place.

Lets keep this PR for updating jira as that's what the bulk of the changes are doing :+1:

Rohlik:
@geekgonecrazy  Okey, I revert category change for Slack-bridge.

geekgonecrazy (approved):

PR #258, opened 2017-03-21
adding Python API

jadolg:


frdmn:
LGTM! You gotta agree with our contributor license first though. 👌

frdmn:
Merged it right away since this a small, non-breaking change

PR #264, opened 2017-03-27
corrected broken link

Prakharsvnit:
@karlprieb

### DIFF
--- a/6. Developer Guides/Code Styleguide/README.md
+++ b/6. Developer Guides/Code Styleguide/README.md
@@ -4,4 +4,4 @@ order: 4
 
 # Rocket.Chat Code Styleguide
 
-- [Less Code Styleguide](https://github.com/Prakharsvnit/Rocket.Chat.Docs/blob/patch-1/6.%20Developer%20Guides/Code%20Styleguide/less.md)
+- [Less Code Styleguide](less.md)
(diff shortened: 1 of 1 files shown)


6. Developer Guides/Code Styleguide/README.md
```diff
@@ -4,4 +4,4 @@ order: 4
 
 # Rocket.Chat Code Styleguide
 
-- [Less Code Styleguide](less)
\ No newline at end of file
+- [Less Code Styleguide](https://github.com/Prakharsvnit/Rocket.Chat.Docs/blob/patch-1/6.%20Developer%20Guides/Code%20Styleguide/less.md)
```
graywolf336: This is not the correct way, as it would just point to your repository and not the web page. Please correct it and make it like other links which are the correct way.

graywolf336 (changes requested):

6. Developer Guides/Code Styleguide/README.md
Prakharsvnit: do you want that link should direct to corresponding web page

graywolf336: No, please see this file for example: https://github.com/RocketChat/Rocket.Chat.Docs/blob/master/6.%20Developer%20Guides/REST%20API/README.md

Prakharsvnit: ok, you want the content itself should be displayed in README.md

graywolf336: No, just the link needs to be corrected to be like the one I pointed you to.

graywolf336 (approved):

PR #265, opened 2017-03-28
Add docs to groups.addAll endpoint and activeUsersOnly param.

nathanmarcos:
Docs to RocketChat/Rocket.Chat/pull/6505

### DIFF
--- a/6. Developer Guides/REST API/Channels/addAll.md
+++ b/6. Developer Guides/REST API/Channels/addAll.md
@@ -49,4 +49,5 @@ curl -H "X-Auth-Token: 9HqLlyZOugoStsXCUfD_0YdwnNnunAJF8V47U3QHXSq" \
 ## Change Log
 | Version | Description |
 | :--- | :--- |
+| 0.55.0 | Added `activeUsersOnly` param |
 | 0.48.0 | Renamed to `channels.addAll` from `channel.addAll` |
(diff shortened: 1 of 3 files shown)


6. Developer Guides/REST API/Channels/addAll.md
```diff
@@ -13,6 +13,7 @@ Adds all of the users of the Rocket.Chat server to the channel.
 | Argument | Example | Required | Description |
 | :--- | :--- | :--- | :--- |
 | `roomId` | `ByehQjC44FwMeiLbX` | Required | The channel's id |
+| `activeUsersOnly` | `true` | Optional <br> Default: `undefined` | Add active users only |
```
graywolf336: Technically it is undefined, but for sake of understanding that the default is false what about changing this to false? Since your code to add this only checks if the value equals `true`?

graywolf336 (changes requested):

6. Developer Guides/REST API/Channels/addAll.md
nathanmarcos: Totally agreed! I'll make the changes.

graywolf336: Awesome, thanks!

graywolf336 (changes requested):
Thanks for making the change, think you can edit the "Change Log" at the bottom of the pages? We can just assume the next version and if it doesn't make it by then, we'll just update it to the correct version.

graywolf336 (approved):
Thanks again! :+1:

PR #268, opened 2017-04-07
Added notes about the custom certificates and SAML Assertion content

failattu:
Hello,

I saw that the pull request for SAML documentation had gone through and would need an enhancement in terms of SAML assertion content and custom cert part. 

This fills in those missing parts.


geekgonecrazy (approved):

geekgonecrazy:
Thanks for the contribution!

PR #273, opened 2017-04-20
Shutdown Rocket.Chat before upgrading

failattu:
The instance of Rocket.Chat should be down while upgrading...


geekgonecrazy (approved):
In this case since you are changing files under it.  Probably recommended.  :+1:. Thanks for your contribution!

PR #276, opened 2017-04-25
Update snap install instructions

tmyoungjr:
How to work around custom kernels of some VPS providers and reference the fact that snapd may need to be installed no matter the version of Ubuntu you have installed


tmyoungjr:
Digging through a bit more.   Going to close this for now - will be back

geekgonecrazy:
Yes lets wait on this.  Please see this conversation i've started here on the snapcraft forum. https://forum.snapcraft.io/t/ubuntu-core-vs-core-snaps/392  I think best to figure out why this is ever necessary, and why this would work

PR #279, opened 2017-04-26
Add snap for desktop

geekgonecrazy:
Add snap to recommended install for desktop


graywolf336 (dismissed):
Maybe alternatively link to the manual install of Mac? Since the store is too outdated

PR #281, opened 2017-04-29
Updated REST API users.list

abrom:
users.list documentation appeared to be a copy of the users.info?? Updated to match actual response. Also added 'query parameters' section with some examples of use


graywolf336:
`v0.49` added the query, sort, and fields. Mind adding that to the history?

abrom:
Done!

What are your thoughts on some of the undocumented API endpoints? Namely:

* users.getAvatar
* users.register
* users.resetAvatar
* users.createToken

Is it a case of it just not having been written yet? Or intentionally undocumented

graywolf336:
Yeah just a case of not being documented for whatever reason. Feel free to add them or I can add them, as I've been working on a decent sized documentation update for the APIs and integrations.

abrom:
No problems. I'll put something together

6. Developer Guides/REST API/Users/list.md
```diff
@@ -9,6 +9,15 @@ Gets all of the users in the system and their information, the result is only li
 | :--- | :--- | :--- |
 | `/api/v1/users.list` | `yes` | `GET` |
 
+## Query Parameters
```
graywolf336: Can you remove the offset, count, and sort ones as they're documented here: https://github.com/RocketChat/Rocket.Chat.Docs/blob/master/6.%20Developer%20Guides/REST%20API/Offset%20and%20Count%20Info.md then put a link to it somewhere in this document? Also, the fields and query ones need a dedicated document to further explain what features are enabled on them.. As the example provided will actually fail due to not being valid json format.

graywolf336 (changes requested):

abrom:
Sorry for the delay @graywolf336. I've updated the docs as requested, and added details to the Offset and Count Info document about the `sort` query parameter

rafaelks:
@graywolf336 Can you review this PR again, please?

graywolf336 (approved):

PR #287, opened 2017-05-09
Add Running under Plesk Onyx behind Nginx section

digitall-it:
As it is a common setup, the specific directives for a standard Plesk Onyx installation are suggested.


geekgonecrazy (approved):

geekgonecrazy:
@digitall-it thanks for sharing your findings here so others will be able to do this :+1:

PR #302, opened 2017-06-06
Add paragraph that explains where the _id is found

cbrummitt:
To help developers figure out where to find the `_id` associated with a certain setting, I added a paragraph that explains that the `_id` is the first argument of the method `add` used in files such as `Rocket.Chat/packages/rocketchat-lib/server/startup/settings.js`. 

In the future, this solution could be augmented by showing a tooltip containing the `_id` whenever an admin hovers over a certain setting.

### DIFF
--- a/6. Developer Guides/REST API/Settings/update.md
+++ b/6. Developer Guides/REST API/Settings/update.md
@@ -3,7 +3,16 @@ order: 20
 ---
 
 # Settings Update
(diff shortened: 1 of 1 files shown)


geekgonecrazy (approved):

geekgonecrazy:
thanks @cbrummitt !

PR #303, opened 2017-06-07
OpenShift: need to use the -p parameter multiple time when creating the app

jfchevrette:
Passing multiple parameters separated by commas is not supported anymore since OpenShift 1.4 released Nov 2016. Ref.: https://github.com/openshift/origin/pull/11539


geekgonecrazy (approved):

geekgonecrazy:
Thanks for the contribution @jfchevrette !

PR #304, opened 2017-06-08
Add instruction to init db

zerotohero:
When first trying to follow this, I was getting errors because the database hadn't been initialized. Add it in to the readme so other users don't run into the same issue.


geekgonecrazy (approved):

geekgonecrazy:
@zerotohero thanks for contributing!

PR #317, opened 2017-06-30
Docs: Explain the concept of webhooks

pixelbrackets:
Explain the existing integrations via WebHooks → Whats the general idea,
which types are available, what is the message format, what are limitations
(recommend to use the REST API for all enhanced features).

I had the same questions while trying for figure out the WebHook concept,
got some answers in the dev chat and thought that this clarification in the docs
may help others as well.

### DIFF
--- a/guides/administrator-guides/integrations/README.md
+++ b/guides/administrator-guides/integrations/README.md
@@ -1,8 +1,19 @@
-# Incoming WebHook Scripting
+# Integrations
 
-With _Scripts_ you can point any WebHook to Rocket.Chat and process the request to print customized messages, define the username and avatar of the _user_ of the message and change the channel where the message will be posted or you can cancel the request to prevent undesired messages.
+Rocket.Chat supports WebHooks to integrate tools and services you like into the platform. WebHooks are simple event-notifications via HTTP POST. This way any application implementing a WebHook is able to POST a message to a Rocket.Chat instance.
 
-## How to create a new Incoming WebHook
+Available integrations:
+
+* Incoming WebHook → send messages from a service into Rocket.Chat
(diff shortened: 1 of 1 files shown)


graywolf336:
While mostly correct, it's not 100%. An outgoing integration can contain a script which does not have to post to a service to create a message in response to an event.

Also, there are several different triggers which can trigger an outgoing integration to be executed not just sending of messages.

Make sense?

pixelbrackets:
Thanks for the feedback :) 

> An outgoing integration can contain a script which does not have to post to a service to create a message in response to an event.

Ah, I see. I'll fix this.

> Also, there are several different triggers which can trigger an outgoing integration to be executed not just sending of messages.

This seems to be a misunderstanding. I wrote that the result of the webhook may be a message  only. The webhook wont trigger any other events, like "create a new channel etc". But the webhook may be triggered by various events. I'll try to clarify this.

pixelbrackets:
I pushed a new version with the fixes mentioned above.

rafaelks:
@graywolf336 Can you review this one again, please?

docsncode:
@graywolf336 are you happy with pixelbracket's changes and ready to merge?

engelgabriel:
@pixelbrackets @SeanPackham @rafaelks I believe this is still not 100% correct. As @graywolf336 explained:

With *Scripts* you can point any WebHook to Rocket.Chat and process the request to print customized messages, define the username and avatar of the *user* of the message and change the channel where the message will be posted or you can cancel the request to prevent undesired messages.

And it is possible possible to trigger any other event with a WebHook (eg. “Create a new channel”).

engelgabriel (approved):

engelgabriel (approved):

frdmn:
Rebased to resolve conflicts after new GitBook structure.

PR #327, opened 2017-07-14
Update minio directions to include options that need set if not using ssl

geekgonecrazy:
ref: https://github.com/RocketChat/Rocket.Chat/pull/7443


rodrigok (approved):

geekgonecrazy:
:+1: thanks!

PR #330, opened 2017-07-18
Improve reporting issue & support docs

geekgonecrazy:
Related to: https://github.com/RocketChat/Rocket.Chat/pull/7492

### DIFF
--- a/1. Contributing/Reporting Issues/README.md
+++ b/1. Contributing/Reporting Issues/README.md
@@ -12,7 +12,6 @@ Examples of things that aren't considered bugs or feature requests:
 
 If your issue isn't a bug please see [Getting Support](/2. Getting%20Support)
 
-
 ## Reporting a bug
 
 Did you find a bug? Reporting is easy!
@@ -22,8 +21,6 @@ Did you find a bug? Reporting is easy!
 3. Try to narrow it down to reproducable steps.  If we can reproduce it, we will be able to fix the issue a lot quicker.
 4. Look at logs for errors. See below for how to gather logs.
 5. If you got this far then: [Open an issue](https://github.com/RocketChat/Rocket.Chat/issues/new)
-
-
 
 ### Things to Include in Bug Report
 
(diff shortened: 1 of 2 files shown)


1. Contributing/Reporting Issues/README.md
```diff
@@ -2,29 +2,52 @@
...
-Here are some suggested things to include, when applicable, which may help us troubleshoot a problem:
+If your issue isn't a bug please see [Getting Support](/2. Getting%20Support)
+
+## Reporting a bug
+
+Did you find a bug? Reporting is easy!
+
+1. [Search the Issues](https://github.com/RocketChat/Rocket.Chat/issues) for your bug.  If you find it, give it a thumbsup and add some additional information to the ticket if differing.  This will help give us a fuller picture.
+2. Make sure you are running the latest version of Rocket.Chat. Its very likely simply updating will resolve your issue.
+3. Try to narrow it down to reproducable steps.  If we can reproduce it, we will be able to fix the issue a lot quicker.
```
gdelavald: > Reproducible

1. Contributing/Reporting Issues/README.md
```diff
@@ -2,29 +2,52 @@
...
+
+Here are some things to include, when applicable, which may help us troubleshoot a problem:
 
 * Operating System / Version / Architecture (64 bit?)
 * Browser type & Version (if web app), any browser add-ons which may be involved (e.g. AdBlocker, NoScript, etc.)
-* Desktop Environment (if desktop app) / Version
-* Rocket.Chat edition / Version or [build number](/2. Getting%20Support)
+* Rocket.Chat Version
 * Expected behavior vs. Actual behavior (In other words, the "bug")
 * Can it be reproduced? If yes, how?
```
gdelavald: I think we can add something like
> Can it be reproduced easily and what are the steps to do it? Or it's something that happens now and then?

1. Contributing/Reporting Issues/README.md
geekgonecrazy: Good catch :+1:

1. Contributing/Reporting Issues/README.md
geekgonecrazy: I think we partially capture that by asking if reproducible.  I'll adjust.  

I think important thing here is we want them to realize for us to fix it.. we need to be able to reproduce the error

gdelavald (approved):

PR #337, opened 2017-07-22
Various changes to the contributing docs, simplification and lint

antgel:


geekgonecrazy:
i'm not in charge of it.  But one that reviews PR's.

The spelling grammer checks here i'm ok with.  The newlines... i'm not a fan of.  

I think that if we have an issue on the docs page with text being to wide, we need to solve it with css not by creating line breaks/ new lines

antgel:
It's not about the text being too wide in the browser. The newlines shouldn't affect display in the browser,  markdown should only treat a *blank* line as a newline.  It's about the considerable pain of long lines when editing.

There are no standards for writing the docs, as far as I can see. Can we get some wider discussion on this? The more fun it is to work on the docs, the better they might become.

geekgonecrazy:
Agreed we do need some guidelines.  Personally i'm not a fan of a sentence just randomly being broke in the middle.  To me it adds more confusion then it being a longer line.

@RocketChat/documentation-and-wiki / @SeanPackham what do you guys think?

antgel:
Well, that's what the full stop is for. 😳

docsncode:
@antgel thank you for your contribution, I think there a lot of useful changes here. Could you split the PR into multiple for each file changed as some will easily be approved and others may need some discussion.

antgel:
@SeanPackham @RocketChat/documentation-and-wiki Unfortunately, my professional situation is going to change in the next few weeks and I'm not sure that I'll have time / licence to work on RC as much as I'd like in that time. I'd really appreciate if somebody could pick this up, because I'm sure much of it would be really useful. Obviously if things change I'll let you know, but it would be a shame to let this go to waste.

PR #344, opened 2017-07-28
Update SAML IdP Attribute info for user mapping

mhou1981:
IdP user attributes mapping requirement is missing so when new user entered the system, they are not shown correct user information.


docsncode:
@mhou1981 the table formatting is broken, could you fix that and then we can get it merged.

mhou1981:
Table fixed.  Just need one extra line before the table syntax started.

engelgabriel (approved):

PR #361, opened 2017-08-03
Sending audio recordings user docs

docsncode:

PR #362, opened 2017-08-04
Sending video recordings

docsncode:
Short snippet on how to send video recordings. (I will be assigning PRs to random people so that the same people don't land up reviewing all of them)


4. User Guides/Editing Messages/README.md
```diff
@@ -53,3 +53,11 @@ Various emoji shortcuts exists to display popular text representations as graphi
 Graphical emojis can be enabled/disabled in [Your Account](../managing-your-account/) settings under the **Messages** section.
 
 When graphical emojis are disabled, emoji keywords and shortcuts will be displayed as is  e.g. `:smile:` or `:)`.
+
+## Sending video recordings
+
+You can send video recordings by clicking on the video camera icon on the right of the compose message box, which will display a live camera preview and recording controls.
```
geekgonecrazy: Right now this is accurate... But This will soon be changed so that accessing is like:

image
image

Just FYI, so we might want an issue for when that does changes so we update this.

geekgonecrazy (approved):
:+1: currently accurate, as mentioned in the comment.  Next version of Rocket.Chat with UI tweaks.. it'll need tweaked

Same thing for #361

PR #372, opened 2017-08-31
Documentation about how to integrate TravisCI with Rocket.Chat

lucasb:
I wrote this script to integrate TravisCI with Rocket.Chat for my project [Fleeg](https://github.com/Fleeg/fleeg-platform). I think it can be useful for others once Travis is a popular CI and is used for many project.


geekgonecrazy (approved):
Thanks for the additions!

PR #376, opened 2017-09-08
Update recommended Node.js version from 4.5 to 4.8.4

### LINKED ISSUE #373: recommended nodejs version
Hi,

which is the recommended nodejs version for RocketChat?

There are for example different information. Here v4.7.1 is recommend:

https://rocket.chat/docs/installation/manual-installation/debian 

Here v4.5.0 is recommend:
https://rocket.chat/docs/installation/manual-installation/ubuntu#2-install-rocketchat

Shouldn't it be version 4.8.4 for security reasons?

Ciao!
Marcus

vcapretz:
The 4.8.4 version is a security release so proably it's a good idea
to recommend users to install this version instead of 4.5 or other ones

Source: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V4.md#4.8.4

closes #373

### DIFF
--- a/3. Installation/4. Manual Installation/CentOS/README.md
+++ b/3. Installation/4. Manual Installation/CentOS/README.md
@@ -44,10 +44,10 @@ Now that we have Node.js and npm installed, we need to install a few more depend
 npm install -g inherits n
 ```
 
-Rocket.Chat needs Node.js version `4.5` using n we are going to install that version:
+The recommended Node.js version for using Rocket.Chat is `4.8.4`. Using _n_ we are going to install that version:
 
 ```
-n 4.5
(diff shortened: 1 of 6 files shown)


geekgonecrazy:
@vcapretz thanks for this.  We need to verify this is supported by the version of meteor we are running.  Its unfortunately not just about recommending the latest, but about making sure we only recommend versions we know are compatible.

vcapretz:
yep @geekgonecrazy, meteor 1.5.1 does support node v4.84: http://docs.meteor.com/changelog.html#v15120170712

engelgabriel (approved):

PR #377, opened 2017-09-08
Unify mongo connection

### LINKED ISSUE #322: Unify suggested mongo connection strings
Ref: https://github.com/RocketChat/Rocket.Chat/issues/7439

vcapretz:
the database name was inconsistent and the docker-compose image names as well
now every mongo connection string should be similar

it closes #322

### DIFF
--- a/3. Installation/2. PaaS Deployments/AWS/README.md
+++ b/3. Installation/2. PaaS Deployments/AWS/README.md
@@ -158,7 +158,7 @@ This guide covers the following:
 2. Create docker-compose.yml, **replacing the ROOT_URL of ABC.DOMAIN.COM with your site**
   `sudo nano /var/www/rocket.chat/docker-compose.yml`
    ```
-   mongo:
+   db:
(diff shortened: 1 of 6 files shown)


geekgonecrazy (changes requested):
I agree on the database name consistency, so thanks for that.  The db name you changed db to mongo but didn't update the links.  I think i'd personally prefer the container name to be `db` as that's what most of us use when we give examples to people just off hand.  So better that the docs stay in tune with that.

vcapretz:
hey @geekgonecrazy thank's for reviewing it! 

So the container name should be `db` and the database I can keep being `rocketchat`?

engelgabriel:
Thanks @vcapretz

I also agree that the container name should be `db` and the database should be `rocketchat`

vcapretz:
done @engelgabriel

geekgonecrazy (approved):

geekgonecrazy:
@vcapretz thanks!

PR #379, opened 2017-09-12
Update login.md

philipbrito:
As said on https://rocket.chat/docs/developer-guides/rest-api#production-security-concerns we should call the REST login api over HTTPS .

### DIFF
--- a/6. Developer Guides/REST API/Authentication/login.md
+++ b/6. Developer Guides/REST API/Authentication/login.md
@@ -18,37 +18,37 @@ order: 1
 
 ## Example Call - As Form Data
 ```bash
-curl http://localhost:3000/api/v1/login \
+curl https://localhost:3000/api/v1/login
      -d "username=myusername&password=mypassword"
 ```
 
 ```bash
-curl http://localhost:3000/api/v1/login \
+curl https://localhost:3000/api/v1/login
      -d "user=myusername&password=mypassword"
 ```
 
 ```bash
-curl http://localhost:3000/api/v1/login \
+curl https://localhost:3000/api/v1/login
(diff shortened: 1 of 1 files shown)


geekgonecrazy (changes requested):
Most people that access localhost like this won't have https.  It would be like 1 in hundreds.  So docs would be better staying with most common.

Also the `\` removed is important for the terminal when copying and pasting the command.

philipbrito:
Not for all OS terminal/bash when copying and pasting:

screen shot 2017-09-21 at 1 15 05 pm

screen shot 2017-09-21 at 1 16 02 pm

philipbrito:
Also, I think that the docs should refer the best approach. If the best approach covers the most of the people then it is nice. For the ones that doesn't have https it could be nice putting an alert in the docs. I belive that most of the people will only replace the username and the password, for the happiness of the sniffers.

What do you think?

geekgonecrazy:
Regarding the `\` since it is multiple lines I agree it should escape the new line for ease of copying and pasting.

As to https.  I strongly disagree with including it.  For the very same reason you gave.  People copy and paste this.  if its localhost:3000 it is **extremely** unlikely https will ever be running on localhost at port 3000.

But for the `\` I agree since it is multi-line. And most likely to work when copy and pasted.

MartinSchoeler:
@filipedelimabrito @geekgonecrazy does this PR needs changes or should be closed?

geekgonecrazy:
Its invalid.  

1. Can't assume https on localhost
2. Can't give command that is multiline with out `\`

So if want copy and paste command needs to be as is

PR #393, opened 2017-09-27
[WIP] Add documentation for the new chat rest api methods

graywolf336:
Would like to have someone review these to make sure none of the copy/paste type errors show their heads...


6. Developer Guides/REST API/Chat/getMessage.md
```diff
@@ -0,0 +1,46 @@
...
+Retrieves a single chat message by the provided id. Callee must have permission to access the room where the message resides.
+
+| URL | Requires Auth | HTTP Method |
+| :--- | :--- | :--- | :--- |
+| `/api/v1/chat.getMessage` | `yes` | `POST` |
+
+## Payload
+| Argument | Example | Required | Description |
+| :--- | :--- | :--- | :--- |
+| `msgId` | `7aDSXtjMA3KPLxLjt` | Required | The id of the message to get. |
```
luciofm, line 15: The documentation seems fine, just that on every other APIs we use `messageId` instead of `msgId`

graywolf336: Sadly these are things we can't change that has already been in Rocket.Chat for several versions now. I have updated the rest api to support `messageId`, but decided to keep it as `msgId` due to older versions.

MartinSchoeler (approved):

PR #396, opened 2017-09-29
Improve documentation

cryptofuture:
I think it could be more common issue. I killed about hour figure out it needs IndexedDB. More above IndexedDB is set to disable in Tor Browser, but it works in it, however doesn't not work with same settings in normal Firefox.
Some commentary on it, welcome. Also, I'm not sure this the best place to add it.


MartinSchoeler:
I think this is a very specific case, the chat not opening could be caused by a lot of different things. Also this should be in a troubleshooting section, not in the users guide.

cryptofuture:
I'll close and send it to troubleshooting.

PR #440, opened 2017-10-31
Update README.md

deenski:
I am not sure if this is something that developers would understand, but I was a touch confused. It's not that difficult to figure out, but for clarity sake, I thought I might propose this.


MartinSchoeler (approved):
That was indeed a typo, thanks for the contribution!

PR #443, opened 2017-10-31
Update docs urls

MartinSchoeler:


geekgonecrazy:
@MartinSchoeler I think no longer needed because of the switch back.

PR #472, opened 2017-11-27
Adding AppVeyor integration for Rocket.Chat

crazy-max:
Here is my contrib linked to issue #471

Issue #471, opened 2017-11-27
AppVeyor integration


crazy-max:
Hi,

I have worked on AppVeyor integration for Rocket.Chat here : https://github.com/crazy-max/rocketchat-appveyor

I have made a PR to add this new integration in the documentation.

Thanks

MartinSchoeler:
Closed by #484

PR #484, opened 2017-12-28
Adding AppVeyor integration for Rocket.Chat

crazy-max:
Here is my contrib linked to issue #471
Uptime Robot integration doc has been updated too with advanced notifications.
Original PR #472 since refactoring.


.gitignore
```diff
@@ -1,4 +1,4 @@
 .DS_Store
-
```
MartinSchoeler: Why did you Change the `.gitignore` file?

crazy-max: To exclude *.iml file (specific to Intellij IDE), is that a problem?

MartinSchoeler: I don't think there is need to add a new exclude to the gitignore file for now.

crazy-max: Ok @MartinSchoeler, i have reverted the commit

MartinSchoeler (approved):
Thanks for the Contribution!

PR #499, opened 2018-01-09
Fix get message http method

MartinSchoeler:
Closes https://github.com/RocketChat/Rocket.Chat/issues/9358


graywolf336 (changes requested):
The curl command needs to be adjusted to display the `msgId` as a query parameter.

graywolf336 (changes requested):
The curl command still needs to be updated

PR #513, opened 2018-01-20
Fixed Blog Links in webpage [updated]

akash-joshi:
Earlier the links used to point to a 404 error page as they were using the old urls


MartinSchoeler (approved):
Thanks for the contribution!

PR #520, opened 2018-01-23
Add tests and update travis.yml

MartinSchoeler:


.travis.yml
```diff
@@ -1,28 +1,16 @@
...
-  - CXX=g++-4.8
-addons:
-  apt:
-    sources:
-    - ubuntu-toolchain-r-test
-    packages:
-    - g++-4.8
+language: ruby
+rvm:
+  - 2.2
```
docsncode: We could bump this to 2.4

Gemfile
```diff
@@ -1,8 +1,10 @@
...
 
-gem 'github-pages', group: :jekyll_plugins
+source 'https://rubygems.org'
+gem 'github-pages', '>=104', group: :jekyll_plugins
 
 group :development, :test do
   gem 'rake'
   gem 'anemone'
+  gem "html-proofer"
+  gem 'mdl', '~> 0.4.0'
```
docsncode: Unless you specifically need this version, you can drop the version and the Gemfile.lock will contain a specific version.

docsncode (changes requested):
Left a couple of comments, looks good.

docsncode (approved):

PR #539, opened 2018-02-04
update user registration API call with correct example parameters

jampod-dev:
The example API call implied that the full registration URL should be included in the payload, when in fact only the secret string appended to the URL should be included.


MartinSchoeler:
@graywolf336 could you take a look at this? Thanks

MarcosSpessatto (approved):

PR #546, opened 2018-02-06
Hard code canonical address

docsncode:
Prevent Jekyll and GitHub pages from using rocketchat.github.io domain


docsncode:
Martin's PR already fixed this. No need for the hard coded URL.

PR #558, opened 2018-02-08
fixed doc about query api

### LINKED ISSUE #557: Mistake in "Query, Fields, and Sort Information" API documentation
The description of 

"fields - json object with properties that have either true or false to include them or exclude them" 

in https://rocket.chat/docs/developer-guides/rest-api/query-and-fields-info/ is wrong. The API does not accept true or false. The correct syntax is "0" for false and "1" for true.

Hence the example is wrong as well.

For additional infos please visit https://docs.mongodb.com/manual/tutorial/project-fields-from-query-results/

janrudolph:
fixes #557


MartinSchoeler:
@graywolf336 is this correct?

MartinSchoeler (approved):

PR #559, opened 2018-02-08
doc: Correct the docs for query api

### LINKED ISSUE #557: Mistake in "Query, Fields, and Sort Information" API documentation

lunaticmonk:
Fixes #557


MartinSchoeler:
@sumedh123 Sorry, but this was already fixed on https://github.com/RocketChat/docs/issues/558 but if you still want to contribute you can find a map with all the missing and outdated docs [here](https://github.com/RocketChat/docs/blob/master/contributing/documentation/documentation-map/README.md). Thanks!

PR #573, opened 2018-02-15
Update README.md to fix numbering typo

dfm066:
Fixing the numbering issue


MartinSchoeler (changes requested):
@dfm066 This pattern seems to break markdown lint tests, maybe try to use an unordered list since there is no need to have a numbered list here. Thanks!

PR #577, opened 2018-02-15
Fix front page links

geekgonecrazy:
Closes #576

PR #581, opened 2018-02-16
Fix absolute link to relative

geekgonecrazy:
Using absolute links only works when serving from local.  Production uses /docs/ so links need to be relative.

ref #577


MartinSchoeler (approved):
LGTM

PR #586, opened 2018-02-19
[DOCS] Add Deeplink documentation

### LINKED ISSUE #497: Deep linking format
Referencing: https://github.com/RocketChat/Rocket.Chat.iOS/issues/440

We need to discuss and document our deep linking format so we can be consistent across all platforms.

Also, discuss which parameters to support so we can have nice deep-link features 🙂

rafaelks:
Closes #497

### DIFF
--- a/developer-guides/deeplink/README.md
+++ b/developer-guides/deeplink/README.md
@@ -1,0 +1,43 @@
+# Deeplink
+
+**This feature is only supported in our iOS native app right now.**
+
+## General information
+
+> rocketchat://
+
+## Authentication
+
+### Without token
+
+**Params:**
+
+- host: The host of the server;
+
+**Example:**
+
+> rocketchat://auth?host=foo.bar.com
+
+### With token
+
+**Params:**
+
+- host: The host of the server;
+- token: The token of the user to be authenticated;
+- userId: The id of the user to be authenticated;
+
+**Example:**
+
+> rocketchat://auth?host=foo.bar.com&token=123abc&userId=1234abcd
+
+## Channel / Group / DM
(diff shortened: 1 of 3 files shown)


developer-guides/deeplink/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+> rocketchat://auth?host=foo.bar.com
+
+### With token
+
+**Params:**
+
+- host: The host of the server;
+- token: The token of the user to be authenticated;
+- userId: The id of the user to be authenticated;
```
cardoso: Mark token and userId as optional. If not provided, we connect to the server, but show login form/methods.

MartinSchoeler (changes requested):
Please add this file to the Table of Contents and Documentation Map.

developer-guides/deeplink/README.md
rafaelks: @cardoso I added two options (with token and without token) but let's make only one, simpler :-)

rafaelks:
@MartinSchoeler How do I do that?

MartinSchoeler:
@rafaelks Just add the file to its respective place in this file https://github.com/RocketChat/docs/blob/master/_data/toc.yml and in this one https://github.com/RocketChat/docs/tree/master/contributing/documentation/documentation-map (just add the file, no need for colored text on this one). You can find more info about the docs map and table of contents in this file https://github.com/RocketChat/docs/tree/master/contributing/documentation/contribution-guidelines

MartinSchoeler (approved):
Looks good to me!

PR #613, opened 2018-02-28
GSOC: Remove Draw (iOS), merge Siri & Apple Watch

cardoso:
We already have the draw feature on iOS.

Siri and Apple Watch were supposed to be a single project.


rafaelks (approved):
Perfect @cardoso 👍

PR #618, opened 2018-03-06
Add better GIF system idea

marceloschmidt:


graywolf336:
Honestly, I'm not sure I agree with this as it is duplicate work for the Rocket.Chat Apps. What is planned for the Rocket.Chat Apps is that we develop, externally, the pop-up which displays the options and use an object notation so that something like `{ url, title, description }` is provided and we show a preview image from the url.

MartinSchoeler:
Closing this since the one that had this idea has given up on it

PR #620, opened 2018-03-07
Clarify that a 'login' operation is required before using Subscriptions.

cmullaparthi:
Clarify that a 'login' operation is required before using Subscriptions.


MartinSchoeler (approved):

MartinSchoeler:
Thanks for your contribution!

PR #650, opened 2018-03-27
Remove windows 10 pro installation guide

MartinSchoeler:
Moved to https://forums.rocket.chat/t/windows-10-pro-installation-guide/408


graywolf336:
What would be the chances a temporary redirect can happen or something similar? That way people who rely on/link to this page can be redirected?

PR #654, opened 2018-03-28
[FIX] Renaming endpoints Get/Post channels.notifications

MarcosSpessatto:
Waiting [PR](https://github.com/RocketChat/Rocket.Chat/pull/10257).
Fix docs affected by renaming endpoints Get/Post channels.notifications to `subscriptions.getOne` and `rooms.saveNotification`.


MartinSchoeler:
@MarcosSpessatto could you add to the respective docs changelogs, in which version were the apis renamed?

graywolf336:
@MartinSchoeler Why? The endpoints were added in 63 and changed in 63, thus no release (and release candidate) shipped with them being how they were previously named. Thus I don't think the version needs to have when it was changed.

MartinSchoeler:
@graywolf336 sorry, you are right, i didn't saw that this files were from 63, reverting now.

MartinSchoeler (approved):

PR #667, opened 2018-04-04
Add admin Restoration doc

### LINKED ISSUE #25: Instructions how to reset lost administration user / password
To quote @marceloschmidt from https://github.com/RocketChat/Rocket.Chat/issues/206#issuecomment-124926626: 

> [...]. If you run with ADMIN_PASS and ADMIN_EMAIL, server will look for an admin user or another user with provided e-mail. If none is found, admin user is inserted. We need to update README accordingly.

MartinSchoeler:
closes #25 

Added guides on how to restore users

### DIFF
--- a/administrator-guides/restoring-an-admin/README.MD
+++ b/administrator-guides/restoring-an-admin/README.MD
@@ -53,6 +53,8 @@ You can either use a one time access token (that will require the user to change
 db.getCollection('users').update({username:"administrator"}, {$set: { "services":{"loginToken":{"token":"some-token-id-that-you-will-use-to-login-once"}}, "requirePasswordChange":true} })
 ```
 
+Then access `http://{your server url}/login-token/some-token-id-that-you-will-use-to-login-once` to login
+
 or you can update the admin password to `12345`
 
 ```javascript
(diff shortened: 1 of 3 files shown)


frdmn (approved):

administrator-guides/restoring-an-admin/README.MD
```diff
@@ -0,0 +1,71 @@
...
+```sql
+use parties
+```
+
+## Updating the admin password
+
+You can either use a one time access token (that will require the user to change his password):
+
+```javascript
+db.getCollection('users').update({username:"administrator"}, {$set: { "services":{"loginToken":{"token":"some-token-id-that-you-will-use-to-login-once"}}, "requirePasswordChange":true} })
```
geekgonecrazy, line 53: You need to make sure to tell them how to actually go and use this.

They need to goto something like: http://{their server}/login-token/some-token-id-that-you-will-use-to-login-once

That will log them in and then let them set the password via the requirePasswordChange dialog

geekgonecrazy (changes requested):

geekgonecrazy (approved):

PR #671, opened 2018-04-05
Better Differentiation Between Quote and Reply

### LINKED ISSUE #665: Add better differentiation between quote and reply
https://github.com/RocketChat/docs/blob/master/user-guides/messaging/README.md#quoting-a-message

Replying a message will trigger a notification to the replied message sender while quoting will not. We should improve the docs to reflect this behavior.

MartinSchoeler:
Closes #665

### DIFF
--- a/user-guides/messaging/README.md
+++ b/user-guides/messaging/README.md
@@ -151,7 +151,7 @@ To copy the contents of a message to your clipboard, simply click on the `copy` 
 
 Replying to a message will create a attachment showing the message you wish to reply after the sent text.
 
-Replying a message will trigger a notification to the replied message sender by using an `@` mention.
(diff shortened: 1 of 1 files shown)


user-guides/messaging/README.md
```diff
@@ -151,10 +151,12 @@ To copy the contents of a message to your clipboard, simply click on the `copy`
 
 Replying to a message will create a attachment showing the message you wish to reply after the sent text.
 
+Replying a message will trigger a notification to the replied message sender by using an `@` mention.
```
geekgonecrazy: > Replying to a message

geekgonecrazy (approved):

PR #689, opened 2018-04-14
Fixing manual install page links.

Slimmons:
On page https://rocket.chat/docs/installation/manual-installation/ all of the links are broken except for CentOS.  I'm assuming this is because they are missing the forward slashes at the end.  It's really a guess, but I can't test it because my dev environment isn't currently up.


MartinSchoeler (approved):

MartinSchoeler:
Thanks for your contribution!

PR #691, opened 2018-04-15
Replace rid with _id

xvonabur:
`rid` is the room id. Room id is the same for the current user.
`_id` is a random id.


MartinSchoeler (approved):
Thanks for the contribution!

PR #693, opened 2018-04-16
[FIX] Fix docs about chat.react endpoint, add parameter to endpoint

MarcosSpessatto:
Fix docs about chat.react endpoint, add parameter to endpoint to make endpoint work like a toggle.
Waiting [PR](https://github.com/RocketChat/Rocket.Chat/pull/10447).


developer-guides/rest-api/chat/react/README.md
```diff
@@ -12,6 +12,7 @@ Toggles the authenciated user's reaction to the provided message.
 | :--- | :--- | :--- | :--- |
 | `emoji` | `smile` | Required | The emoji to react with. |
 | `messageId` | `7aDSXtjMA3KPLxLjt` | Required | The message id to react to. |
+| `shouldReact` | `7aDSXtjMA3KPLxLjt` | Optional: `true` or `false` | Parameter to make endpoint work like a toggle. |
```
cardoso: The way it worked before was like a toggle. "shouldReact" makes it work like a setter.

I think it would be better explained this way:

```
Parameter to specify if the reaction should be added or removed
```

developer-guides/rest-api/chat/react/README.md
```diff
@@ -37,5 +38,6 @@ curl -H "X-Auth-Token: 9HqLlyZOugoStsXCUfD_0YdwnNnunAJF8V47U3QHXSq" \
 
 | Version | Description |
 | :--- | :--- |
+| 0.64.0 | Added parameter "shouldReact" to make endpoint work like a toggle |
```
cardoso: `work like a setter`

cardoso (changes requested):
Nice work with the API @MarcosSpessatto 

I think we can explain this better:

PR #705, opened 2018-04-19
Update README.md

ZsZs73:
Add python-dev as build dependency otherwise `npm install` will fail with `ImportError: No module named compiler.ast` while doing `node-gyp rebuild` in `dtrace-provider` on Debian 9 Stretch
Credit should go to https://github.com/lovell/sharp/issues/1087


MartinSchoeler (approved):
Thanks for the contribution!

PR #725, opened 2018-05-01
Rocketchat download link broken

STNN:
https://releases.rocket.chat/latest/download DOES NOT work anymore, spent a half-hour trying to figure out what i was doing wrong - I am a linux newbie and have no clue what i am doing until realized that this is the proper link
https://download.rocket.chat/stable

Please save other noobs the same headache


engelgabriel:
Hi, thanks for the help, but the original link is now fixed.

PR #763, opened 2018-05-24
Update snap restore intructions

kevincolten:
- Formatting was broken in a few places
- Must complete Initial Setup Wizard first
- When restoring database, index clash will break restore, must use `--noIndexRestore`


MartinSchoeler:
@kevincolten Thanks for the contribution, but those changes break the markdown linting, what you can do is instead of using headers for the notes (that now use headers), use a quote `>` that will give a way better formatting and will not break the linting. you can find our markdown rules in [here](https://rocket.chat/docs/contributing/documentation/markdown-styleguide/#markdown-style-guide). Thanks!

engelgabriel:
Thanks @kevincolten for the contribution, but we have updated the doc tu use 
```
sudo snap run rocketchat-server.restoredb /var/snap/rocketchat-server/common/rocketchat_backup.tgz
```

PR #777, opened 2018-06-01
[NEW] Add permissions.list and permissions.update, and deprecate permissions

vynmera:
Waiting [PR](https://github.com/RocketChat/Rocket.Chat/pull/10975).

This PR adds the aforementioned new features to the docs. This **deprecates** `/api/v1/permissions`, so I added it to the deprecated features. However, to make place for an index, I had to remove the original page for it, if there is any place I can re-add it please let me know!

Since I do not know yet what version it will be merged in (if at all), I entered 0.66.0 for now (and 0.69.0 for the deprecation). If this is not the case, please change it or request a change.


MartinSchoeler (approved):
This looks good to me. Thanks for the contribution!

PR #799, opened 2018-06-18
Change [your_app].slack.com

tjhorner:
`my.slack.com` redirects to whatever the primary workspace of the signed in user is, so they don't need to fill it in themselves.


MartinSchoeler (approved):

MartinSchoeler:
Thanks for the contribution!

PR #823, opened 2018-07-04
Remove nano installation in CentOS

tassoevan:
Prefer the default `vi` over `nano` as text editor.


tassoevan:
I've found more room for improvements, following [this](https://forums.rocket.chat/t/proposal-for-officially-supported-installation-methods/405).

PR #829, opened 2018-07-09
Update hubot image and bot account creation.

robindijkhof:
Use latest hubot image as 0.1.4 won't work with the current versions of Rocket.Chat.
Use docs from RocketChat/hubot-rocketchat from bot account creation.


MartinSchoeler (approved):

MartinSchoeler:
Thanks for the contribution!

PR #836, opened 2018-07-13
Update Documentation: Channel Kick

epaaj:
Fixed use of wrong word in payload description.

PR #838, opened 2018-07-13
Clarifying Documentation for Group Kick

faelwar:
Change description of userId to "The user id of who to remove" instead of "..who to invite"


MartinSchoeler:
Already done in https://github.com/RocketChat/docs/pull/836 but thanks for having interest on Rocket.Chat docs.

PR #853, opened 2018-07-24
Fix Documentation: Rest API > User list

rogeriopradoj:
Closes #852 - Change overflow to auto in .wrapper

### DIFF
--- a/_sass/layout.scss
+++ b/_sass/layout.scss
@@ -149,7 +149,7 @@ blockquote {
   max-width: $content-width;
   margin: 0 auto;
   flex: 1 1 100%;
-  overflow: hidden;
+  overflow: auto;
 
   &.main {
     display: flex;


rogeriopradoj:
@diegodorgam hello! I'm Rogerio, one of users of rocketchat in https://chat.caixa.gov.br/ (https://chat.caixa.gov.br/direct/c066770). See ya!

MartinSchoeler:
Unfortunately this will create another problem, it will make the whole main content element scroll if one table overflows the main content width. That doesn't look right. To fix this we need make the table scrollable. This can be done by adding the table to a `<div>` element with `overflow-x: auto`, since jekyll generates automatically the tables, it has to be added via js. I've already done this in #854. But thanks of the interest in contributing to Rocket.Chat documentation!

PR #855, opened 2018-07-26
Facebook strict mode for redirect urls

### LINKED ISSUE #11389: OAuth with Facebook not working: Can't Load Url The Domain of this url Isn't Included in the app's domain
Facebook shows the error that there is no URL included in the app domain setting although there are all domains and subdomains set.
bildschirmfoto von 2018-07-08 19-49-18
I suspect that its because rocket chat still uses the 2.12 API. In the new API you have to use strict redirect URIs and perform the call like this:
`https://www.facebook.com/v3.0/dialog/oauth?
  client_id={app-id}
  &redirect_uri={"https://www.domain.com/login"}
  &state={"{st=state123abc,ds=123456789}"}`
but the call is like this:
`https://www.facebook.com/v2.12/dialog/oauth?client_id=1234&redirect_uri=https://www.domain.com/_oauth/facebook?close&display=popup&scope=email&state=abcabcalotofsecretletters=`
therefore it becomes impossible to enter all the possible redirect URIs.

I use 0.66.2 on Ubuntu Server 18.04 LTS. I installed rocket chat via snap

fspc:
Facebook now enforces strict mode for redirect urls, <<website_url>>/_oauth/facebook?close is now the proper format.
Fixes https://github.com/RocketChat/Rocket.Chat/issues/11389

### DIFF
--- a/administrator-guides/authentication/oauth/README.md
+++ b/administrator-guides/authentication/oauth/README.md
@@ -6,7 +6,7 @@ These settings are in the `Accounts` setting page under `Administration`.
 
 ## Facebook
 
-- Callback url: `<<website_url>>/_oauth/facebook`
+- Callback url: `<<website_url>>/_oauth/facebook?close`
 
 ## GitHub
 


MartinSchoeler:
Hey @fspc could you sign the CLA?

fspc:
@MartinSchoeler CLA signed.

MartinSchoeler (approved):

PR #856, opened 2018-07-26
Clarify that channel.* methods are for public channels.

bachmeil:


MartinSchoeler (approved):
This looks right to me. Thanks for the contribution!

PR #865, opened 2018-08-01
Create LICENSE.md

jamalv:
Added MIT licence, the same licence used for rocket.chat itself.

Motivation: we’d like to use the Jekyll theme (looks great) and were wondering if that’s okay :)


MartinSchoeler:
@jamalv Thanks for adding that. But if you are going to use the theme, i would suggest getting it from https://github.com/RocketChat/rocketchat.github.io since the docs uses less styling than the full website.

MartinSchoeler (approved):

PR #897, opened 2018-08-31
Add others way to install nodes and autostart rocketchat service

zhaojul:
Add others way to install nodes and autostart rocket chat service, I used yum mirror install the nodes,simplified the related steps, but the results are the same。

### DIFF
--- a/installation/manual-installation/centos/README.md
+++ b/installation/manual-installation/centos/README.md
@@ -51,6 +51,15 @@ The recommended Node.js version for using Rocket.Chat is `8.9.3`. Using _n_ we a
 n 8.9.3
 ```
 
+**HELP**
+
+There is also a way to install nodejs：
+
+```
+curl --silent --location https://rpm.nodesource.com/setup_8.x | bash -
+yum -y install nodejs
+```
+
 ## Installing Rocket.Chat
 
 Now we download and install Rocket.Chat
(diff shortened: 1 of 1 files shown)


zhaojul:
Add others way to install nodes and autostart rocket chat service, I used yum mirror install the nodes,simplified the related steps, but the results are the same。

geekgonecrazy:
Thanks for the contribution.  Looks like everything would work fine.  Only concern is adding more variables to the recommended way to set it up.  

The Node.js part is legit, and possibly what we should use as the recommended way.

Adding supervisor in the mix though... I do question if we want to include that in the docs.  Because if we include people will start using it.

What does supervisor specifically bring to table that is better then a systemd unit file?

zhaojul:
Forgive my English is not good, I use the supervisor to start the rocket.chat under centos6, add the official service startup file in the use of centos7, but the service under centos7 can not start itself after booting, and the user's mailbox verification function can not be Use, so I wrote a simple script, using the script to complete the startup and successfully send the user's mailbox verification email. Adding the supervisor is to provide other solutions when the service file is not started properly.

geekgonecrazy:
I think we do not officially support centos 6.  I don't like the idea of including another way to daemonize.  We prefer to standardize on systemd since now supported by all OS's

Maybe we can add in a Centos 6 guide in the community guides?

geekgonecrazy:
Thanks for your time on this and for using Rocket.Chat.  We want to keep the number of possible ways to install Rocket.Chat down to a managable level for us to be able to support.  That way we can make sure those work great, and every one will benefit.

I think this would be better suited as a post here in the community guide section of the forum: https://forums.rocket.chat/c/community-guides

Please feel free to create as a post there for CentOS 6.

PR #900, opened 2018-09-03
Add pricing links to docs site

MartinSchoeler:
This adds the new pricing page links to the header and footer of the docs site


theorenck (approved):
It seems to be ok

PR #927, opened 2018-09-27
Add/e2e doc

geekgonecrazy:


MartinSchoeler (changes requested):
Can you add this file to the table of contents?

MartinSchoeler:
Also it might be nice to have a link to this page in here https://github.com/RocketChat/docs/blob/master/user-guides/README.md

MartinSchoeler:
You can also create a new type of channel in [this](https://github.com/RocketChat/docs/tree/master/user-guides/channels) article and link it to the e2e article

MartinSchoeler (approved):

MartinSchoeler:
Looks good to me!

geekgonecrazy:
Thanks!  🤗

user-guides/end-to-end-encryption/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+Once you have this password saved somewhere it is safe to dismiss the alert.
+
+## Changing Encryption Password
+
+If you would like to change your encryption password you can do so by going to: My Account->Encryption
+
+![ ](e2e-changepassword.png)
+
+Once you make this change you will need to update the password on any other devices being used.
```
rodrigok, line 25: This part is not true, the user will need the new password only when importing the credentials again, usually in a new client

geekgonecrazy: ah.. good to know! i'll update

PR #929, opened 2018-09-28
Make format of SAML Cert and Key more clear

pkolmann:
I just setup rocket.chat with SAML and fell for this hole. Seems that others have hit this wall as well.

https://github.com/RocketChat/Rocket.Chat/issues/9634
https://github.com/RocketChat/Rocket.Chat/issues/6317

Please commit this change to make it more visible how to use the data.
Thanks
Philipp


MartinSchoeler (approved):
Thanks for your contribution!

PR #932, opened 2018-10-02
Updated for CentOS 7.5

MikeD096:
Added updates for CentOS 7.5, mongodb 4.0, and Node.js 8.11.3,


installation/manual-installation/centos/README.md
```diff
@@ -19,12 +19,12 @@ nano /etc/yum.repos.d/mongodb-org.repo
 Paste this into the new file:
 
 ```
-[mongodb-org]
+[mongodb-org-4.0]
```
geekgonecrazy, line 22: We aren't actually ready to suggest 4.0 since its not officially supported by the framework we use

geekgonecrazy:
Thanks for your contribution.  One main issue with this is the mongo bit.  4.0 is not currently one of the versions we recommend.

installation/manual-installation/centos/README.md
LuluGO: Hi, apart from the mongo section which should stay with version 3.6, I notice you added some commands using sudo but the rest of the guide is without using sudo and that could be confusing, maybe remove the sudo? Thank you! Great change for the nodejs installation.

LuluGO (commented):
Remove 'sudo' as it could be confusing, the rest is great

installation/manual-installation/centos/README.md
MikeD096: The sudo was added as I was running into errors for those commands without it, even running as root.  I will spin up a clean VM this weekend and retest to verify.

bkmgit:
Thanks for the update. May also want to look at the guide [here](https://linuxize.com/post/how-to-deploy-rocket-chat-on-centos-7). One issue that may also arise when using Nginx proxy is requiring to change selinux permissions.

MikeD096:
Closing as doc has since been overhauled by @LuluGO and my changes are no longer required.

PR #951, opened 2018-10-15
Update README.md

reetp:
Remove install instructions for snaps from this page and point the user to the full Snap installation page which also has logging and debugging information.

There should really be a 'Manual installation' subdirectory/menu entry with the Manual information on this page moved to a sub page which could also include debugging information.


MartinSchoeler (approved):

MartinSchoeler:
Thanks for the contribution!

PR #953, opened 2018-10-16
[NEW] Improve 2fa docs

MartinSchoeler:
This adds more information to the existing 2fa docs.


engelgabriel:
@MartinSchoeler where are the screenshots?

MartinSchoeler:
@engelgabriel It was decided that the docs shouldn't use many screenshots, especially when the subject of the article is not that hard to find. Adding many screenshots can make very difficult to maintain them.

PR #954, opened 2018-10-16
Update Documentation: REST API (users.setAvatar)

samsimpson1:
Update the setAvatar docs to add the image argument to the arguments list, and change the wording of the avatarUrl argument.


MarcosSpessatto (changes requested):
I think you should explain that the `image` parameter is only sent by `form-data`, because the way you put it, it looks like it might be sent by the `POST` request payload. What do you think of putting a new section describing which fields should be sent if the user wants to upload the image(send by `form-data`)?

samsimpson1:
I've split the page into two, one for when you want to upload a file and one for when you're specifying a URL and clarified that the image needs to be sent as form data.

MarcosSpessatto (approved):

MarcosSpessatto:
@TopHattedCat  thanks! :+1:

PR #957, opened 2018-10-23
Update README.md

reetp:
The tar extract instruction does mv /tmp/bundle /tmp/Rocket.Chat and then cd /tmp/bundle/programs/server which is to a directory that then no longer exists.

Updated to match the correct instruction from the Debian Manual Install doc.


MartinSchoeler (approved):

MartinSchoeler:
Thanks for your contribution!

PR #960, opened 2018-10-29
Adding README.md file for Authentication

Sagar2366:


Sagar2366:
Please suggest modification you need for this doc. I would like to work on this.

MartinSchoeler:
It was only a matter of changing the list styling, fixed that and it looks good. Thanks for the contribution!

MartinSchoeler (approved):

PR #974, opened 2018-11-03
German User Group channel

localguru:
Note on the channel of the German Rocket.Chat User Group, founded with @TwizzyDizzy

We think it's a good idea to check back with RC folks like @tostanoski or @frdmn. Maybe we can also have a look for already existing groups that we could rename to the ug_[language] format.

Any feedback, guys?

Cheers
Thomas & Marcus


MartinSchoeler (approved):

MartinSchoeler:
I think this would be a good addition to the docs, for now let's merge this and create a new PR if we find/create new channels. Thanks for the contribution!

PR #985, opened 2018-11-13
[New Documentation] Added E2E methods

Hudell:


rafaelks:
@Hudell Can you please add the minimum version required to use this APIs?

Hudell:
I added a changelog at the end of each file with the version the methods were included on.

rafaelks (changes requested):
@Hudell Thank you! The only thing now is that you only added the documentation for the real-time methods... can you also add it to the REST APIs please?

engelgabriel (approved):

PR #988, opened 2018-11-15
Update README.md

axelolmos:
graphicsmagick is actually GraphicsMagick
npm install as root seems to fail but doing it as user rocket chat works.


LuluGO:
npm install as root fails, but the guide was meant to be run as a non root user, that is why is written with sudo, did you try to run it as root?
You are right about GraphicsMagick, also epel repo should be installed, thanks!

axelolmos:
Sorry, I ran it as root.

KickStart scripts generally run as root when installing software and su to other users, so I was coming from the opposite way of thinking.

Thanks,
Axel

LuluGO:
Closing, and fixed the GraphicsMagick with epel install in https://github.com/RocketChat/docs/pull/1002
Thanks!

PR #993, opened 2018-11-21
Add client_max_body_size to 200M to the documentation as basic setup

rafaelks:


MartinSchoeler (approved):

geekgonecrazy:
Might not hurt to include a note somewhere to increase if file limit is higher.  I know some servers that have 1G as max

PR #998, opened 2018-11-22
[NEW] Improve quick start page

reetp:
Added more detail on setting up a development environment on Linux
It would be good if someone tested this on a clean machine

### DIFF
--- a/developer-guides/quick-start/README.md
+++ b/developer-guides/quick-start/README.md
@@ -15,12 +15,13 @@ Install Meteor:
 `curl https://install.meteor.com/ | sh`
 
 Notes here:
+
 - [Git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git)
 - [Meteor](https://www.meteor.com/install)
 
 Install latest npm:
 
-npm i npm@latest -g
+`npm i npm@latest -g`
 
 Add 'n' so you can easily swap node versions:
 
--- a/developer-guides/quick-start/README.md
+++ b/developer-guides/quick-start/README.md
@@ -8,10 +8,11 @@ To run Rocket.Chat for development on a Linux or a Mac you must first have the f
 
 These will require root privileges:
 
-Install g++ build-essential git mongodb
+Install `g++ build-essential git mongodb`
 
 Install Meteor:
-curl https://install.meteor.com/ | sh
+
+`curl https://install.meteor.com/ | sh`
 
 Notes here:
 - [Git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git)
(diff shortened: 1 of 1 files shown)


MartinSchoeler (approved):

MartinSchoeler:
Looks good to me.

fyngyrz:
@reetp Suggestion: rather than saying "these will require root privileges", and then (this is missing) saying elsewhere what does NOT require root privs, I suggest you write the install as...

**sudo** apt... etc.

...where required. Either that, or add a line at the appropriate juncture(s) that says "These do not require root privs:"

These instructions need to be tested in a new VM with _no_ assumptions made. Don't rely on what you, as a doc author, know. Presume it's all new to the person trying to do the install, and write the instructions so that the install will succeed. Every step needs to be documented. Some developers (including me) will be new to the ecosystems running here: npm, meteor, monogodb, etc., possibly even linux (for instance, the install recommends that Windows users install in a linux VM... Okay, but don't make that even harder on them. They may already be facing an uphill challenge just because of being new to linux.) The install docs should not presume familiarity before the developer even starts looking at the system.

If a dev is familiar with some or all of this, the explicit level of detail will do no harm. If they are not, it is fundamental to a successful install.

I define a "successful install" as one that actually results in the "SERVER RUNNING" block as described at the end of the `Quick start for code developers` section. Leaving out _anything_ that is required to get to that point in a new VM or linux install is a documentation shortcoming.

My background, FWIW, is as a desktop application and Python developer. I've been at this since the 1970's, and have lots of successful projects in those areas to my credit on both Windows and OS X, including having written a large amount of documentation for same. I can get around linux somewhat, but outside of c++ and Python, and various markup languages (HTML, CSS) I'm starting fresh. I'm "that guy" when it comes to your linux developer install: pretty much a blank slate.

So I hope you'll take my input on the install procedure seriously. Thanks for reading.

**Please use an `@fyngyrz` when replying to me so I will be notified.**

fyngyrz:
@reetp In the quickstart:

> `Add 'n' so you can easily swap node versions:`
> `n 8.11.4 (use one similar to the official builds)`

What does this mean? Entering `n 8.11.4` at the command prompt results in `n command not found`

`add 'n'` needs to be explicit, in other words, `sudo apt install  n` or whatever is actually needed here. Also , "so you can easily swap node versions" explains nothing. What's a "node version"? Why do I need to easily swap it?

If `n`, whatever it is, actually needs to be there for something we will run into later, just provide the install step without comment. If it's optional, which seems to be implied here, either fully explain why it is optional so an intelligent decision can be made, or drop it from the install docs. Again, this problem arises because the install docs presume an understanding that may very well not be present.

**Please use an `@fyngyrz` when replying to me so I will be notified.**

fyngyrz:
@reetp FYA:

So far, in a new Ubuntu 18.04.1 VM, the following install steps, in this order, seem to actually be what ought to be in the install instructions up to the completely opaque `add 'n'` remark...

```
sudo apt install g++
sudo apt install build-essential
sudo apt install mongodb
sudo apt install git-all
sudo apt install curl
curl https://install.meteor.com/ | sh
sudo apt install npm
npm i npm@latest -g
```

...I will wait on proceeding until I know what to do, if anything, about `add 'n'`

However, I think the above needs to be added to the install instructions (so far.)

**Please use an `@fyngyrz` when replying to me so I will be notified.**

PR #1011, opened 2018-12-04
Added LD_PRELOAD to service script

gerricom:
Current CentOS releases only ship with zlib 1.2.5 or older. Sharp needs a more current build of this, at least 1.2.9, otherwise this will lead to an exception as reported here: https://github.com/RocketChat/Rocket.Chat/issues/12788
Luckily a recent build is included to sharp's node_module so you can `LD_PRELOAD` this.


MartinSchoeler (approved):

MartinSchoeler:
Thanks for your contribution!

Issue #1021, opened 2018-12-09
Typo in developer guides internationalization section


pranay414:
In [https://rocket.chat/docs/developer-guides/internationalization/](https://rocket.chat/docs/developer-guides/internationalization/) there is a typo in introductory paragraph `When developing in Rocket.Chat, you have the ablilty to add strings to our tranlations files in wich later will be tranlsated to other languages by our community translators.` It should be `translations files`.

MartinSchoeler:
Closed by #1022

PR #1022, opened 2018-12-09
[FIX] Typos in developer guides internationalization section

pranay414:
Fixes typo in developer guides internationalization section and issue #1021


MartinSchoeler (approved):
Looks good to me, thanks for the contribution!

PR #1035, opened 2019-01-08
Update README.md command should be: npm run postinstall

barak-haviv:


d-gubert (approved):

d-gubert:
Thank you for the contribuition! :smiley:

PR #1086, opened 2019-02-09
Update readme.md

Rodriq:
An update on the *README.md* file, simpler instructions.
  Providing an alternative command in case of bundle version mismatch.


Rodriq:
Oh sorry I remember accepting that.
  Let me review one  moment.

PR #1110, opened 2019-02-26
Fix API call: get by query parameters

adpe:
Instead of get the response by json array

### DIFF
--- a/developer-guides/rest-api/groups/counters/README.md
+++ b/developer-guides/rest-api/groups/counters/README.md
@@ -6,7 +6,7 @@ Gets group counters.
 | :--- | :--- | :--- |
 | `/api/v1/groups.counters` | `yes` | `GET` |
 
-## Payload
+## Query Parameters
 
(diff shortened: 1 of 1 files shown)


MarcosSpessatto:
@adpe sorry but I could not see the difference by removing the json header. :thinking:

MarcosSpessatto (commented):
@adpe sorry but I could not see the difference by removing the json header. thinking

adpe:
> @adpe sorry but I could not see the difference by removing the json header. thinking

Mhm - you don't pass in this call the data by json as you set the parameter by the URL. So, this header is not necessary? In my environment it was not working setting this header.

MarcosSpessatto:
Weird, it should works with the header as well. Can you please add your change to the `channels.counters` and `im.counters` also?

adpe:
Hi @MarcosSpessatto 

Just tested it out again, it's working now. I think I've done some mistake using this last time.

PR #1113, opened 2019-03-02
Reverse Proxy with Let's Encrypt Doc

clarm060:
- Update centos Manual Install page terminology
-  Adding SSL Reverse Proxy with Let’s Encrypt doc
- added linkage on all needed pages for new Let's Encrypt page.


MartinSchoeler (changes requested):
Can you add the new file to the [Table of Contents](https://github.com/RocketChat/docs/blob/master/_data/toc.yml)? Thanks!

Faria1212 (approved):

PR #1123, opened 2019-03-06
Add docs for federation

sampaiodiego:
Related to https://github.com/RocketChat/Rocket.Chat/pull/12370/


administrator-guides/federation/README.md
```diff
@@ -1,11 +1,88 @@
 # Federation
 
+## Introduction
+
+Federation allows communication with other Rocket.Chat servers.
+
+When you open Federation's config screen, you will see this:
+
+![Initial Federation Screen](./initial.png){:style="width='100%' height='auto'"}
```
sampaiodiego: seems `{:style ...}` isn't a valid GitHub Markdown, as seen [here](https://github.com/RocketChat/docs/blob/f7748c67de2cf80336c275f0e96df8174459fe58/user-guides/federation/README.md)

but on GitHub (and hopefully our website) that's the default style for images =) so we can remove all styles.

administrator-guides/federation/README.md
```diff
@@ -1,11 +1,88 @@
...
+- Weight: `1`
+- Priority: `1`
+- TTL: `1 hour`
+- Target: `chat.mydomain.com`
+- Port: `80`
+
+#### TXT Record
+
+- Host: `rocketchat-public-key`
+- Value: `<my public key, as shown on the configuration screen>`
```
sampaiodiego, line 101: should we add an special tip for Route53 users regarding need to split the key in two string?

user-guides/federation/README.md
```diff
@@ -0,0 +1,32 @@
...
+- Username: `myfriendsusername@anotherdomain.com`
+- E-mail address: `joseph@remotedomain.com`
+
+If the remote server is within Rocket.Chat Federated Network, you will be able to see the users:
+
+![Directory Search Result](./directory-result.png){:style="width='100%' height='auto'"}
+
+## Inviting remote users and chatting
+
+If you click on one of those listed users, you will invite them to your server without login access. Also, you and everyone else on your server, will be able to chat with that new remote user.
```
sampaiodiego:
```suggestion
If you click on one of those listed users, you will invite them to your server without login access. Also, you and everyone else on your server will be able to chat with that new remote user.
```

administrator-guides/federation/README.md
```diff
@@ -0,0 +1,88 @@
+# Federation
+
+## Introduction
+
+Federation allows communication with other Rocket.Chat servers.
```
Sing-Li: Can we please have a small introduction /  architecture section explaining things such as what the Hub is, why is DNS relevant to Federation and and how can servers "trust" one another to get Federation going  (can we just force impose Federatation to any other server anywhere ?  that is not clear at all)

administrator-guides/federation/README.md
```diff
@@ -0,0 +1,88 @@
...
+- **Disabled :** federation is disabled
+- **Running :** federation is running and ready to communicate with other servers;
+
+#### Domain
+
+Fill this option with your domain name, usually where the Rocket.Chat server is hosted.
+
+#### Public Key
+
+All Federation communications are double encrypted, which means messages cannot be delivered by mistake to a wrong peer. This is your public key, which is specially generated for Rocket.Chat and by Rocket.Chat, which will be shared with peers you are communicating with.
```
Sing-Li, line 54: Is this totally true?  do we mean that it cannot be decrypted even if it is misdelivered?  Or do we decrypt verify and then refuse to accept that message if it is not for us?

administrator-guides/federation/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+#### Hub URL
+
+This usually remains as is. This is the Hub URL, where your server will register and confirm domain ownership if you are not using DNS discovery method.
+
+#### Discovery Method
+
+How are you going to be found in the Rocket.Chat Federated Network? You can choose through DNS (recommended) or Rocket.Chat Hub.
```
Sing-Li: Why would we choose one over the other?  Please explain.

administrator-guides/federation/README.md
```diff
@@ -0,0 +1,88 @@
...
+## Introduction
+
+Federation allows communication with other Rocket.Chat servers.
+
+When you open Federation's config screen, you will see this:
+
+![Initial Federation Screen](./initial.png){:style="width='100%' height='auto'"}
+
+Below, a brief explanation of each configuration option:
+
```
Sing-Li, line 32: Need a high level description of *what* is implemented in this version, what should users be expecting to be working.   And what are the known restrictions.  What will be coming in future versions.

Sing-Li (changes requested):

administrator-guides/federation/README.md
alansikora: The encryption process encrypts the message using the Server A private key, and the Server B public key.

When a server receives the message, it uses the Server B private key and the Server A public key to decrypt it.

So, let's say the message ended up on Server C: this server won't have Server B's private key to decrypt it, so the message can't be read at all.

administrator-guides/federation/README.md
alansikora: Cool!

administrator-guides/federation/README.md
alansikora: @sampaiodiego I thought about it, but there are so many different DNS managers out there, it is going to be hard to, eventually, add instructions for all of them.

I thought about saying that some DNS providers won't allow the full string to be entered, so after saving, we recommend making sure the string is complete and if it is not, the string needs to be entered in more than one key, what do you think?

alansikora:
@Sing-Li I've made some changes, can you please take a look at those? Thanks for the suggestions!

MartinSchoeler (changes requested):
Please don't forget to add the new files to the [Table of Contents](https://github.com/RocketChat/docs/blob/master/_data/toc.yml)

administrator-guides/federation/README.md
sampaiodiego: I like it, I think it is enough saying that way

administrator-guides/federation/README.md
geekgonecrazy: Can we put that into words here in the doc?  Almost exactly as is would be good.  I think this would help others understand the encryption and how its secure also

user-guides/federation/README.md
```diff
@@ -0,0 +1,32 @@
+# Talking to users from another server
+
+Once you have [added your server to the Federated Network](../adminitration-guides/federation) you can talk to users from server who have done the same.
```
geekgonecrazy:
```suggestion
Once you have [added your server to the Federated Network](../administrator-guides/federation) you can talk to users from server who have done the same.
```

administrator-guides/federation/README.md
```diff
@@ -0,0 +1,110 @@
...
+#### SRV Record
+
+- Service: `_rocketchat`
+- Protocol: `_tcp`
+- Name: `mydomain.com`
+- Weight: `1`
+- Priority: `1`
+- TTL: `1 hour`
+- Target: `chat.mydomain.com`
+- Port: `80`
```
geekgonecrazy, line 93: maybe embed an image like this? 

image

administrator-guides/federation/README.md
geekgonecrazy: This should be made more clear as it reads now... I would want to create a dns record at rocketchat-public-key.mydomain.com

I know this is implementation related.. but in my opinion we should do a subdomain for the txt record.  Most orgs already have spf taking up a TXT on their root domain

administrator-guides/federation/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+#### TXT Record
+
+- Host: `rocketchat-public-key`
```
geekgonecrazy:
```suggestion
- Host: `@.mydomain.com`
```

administrator-guides/federation/README.md
```diff
@@ -0,0 +1,110 @@
...
+- Value: `<my public key, as shown on the configuration screen>`
```
geekgonecrazy, line 101:
```suggestion
- Value: `rocketchat-public-key=<my public key, as shown on the configuration screen>`
```

administrator-guides/federation/README.md
```diff
@@ -0,0 +1,110 @@
...
+- **Disabled :** federation is disabled
+- **Running :** federation is running and ready to communicate with other servers;
+
+#### Domain
+
+Fill this option with your domain name, usually where the Rocket.Chat server is hosted.
+
+#### Public Key
+
+All Federation communications are double encrypted, which means messages cannot be delivered by mistake to a wrong peer. This is your public key, which is specially generated for Rocket.Chat and by Rocket.Chat, which will be shared with peers you are communicating with.
```
geekgonecrazy, line 54:
```suggestion
All Federation communications are double encrypted, which means messages cannot be delivered by mistake to a wrong peer. This is your public key, which is specially generated for Rocket.Chat and by Rocket.Chat, which will be shared with peers you are communicating with.

The encryption process encrypts the message using the Server A private key, and the Server B public key.

When a server receives the message, it uses the Server B private key and the Server A public key to decrypt it.

So, let's say the message ended up on Server C: this server won't have Server B's private key to decrypt it, so the message can't be read at all.
```

administrator-guides/federation/README.md
geekgonecrazy, line 93:
```suggestion
- Port: `80`

In route 53 it would look something like this:
![image](https://user-images.githubusercontent.com/51996/53998274-c1496c80-4104-11e9-8571-3288fc7b2cc8.png)
```

sampaiodiego:
Added do ToC and also do "projects" (initial) page. It should be good to merge. Please review

administrator-guides/federation/README.md
alansikora: @geekgonecrazy is TXT a single record?

administrator-guides/federation/README.md
alansikora: I like that

administrator-guides/federation/README.md
geekgonecrazy: For route 53 yeah its one big ugly record :grin: 

Its not the end of the world.. but I know for DKIM most of the time they have some dedicated TXT record just for it and nothing else

alansikora: Yeah, I just saw that 🤦‍♂️ 

We could change that, should be easy. @sampaiodiego what do you think?

alansikora: @geekgonecrazy we are going to make that change, thanks for bringing that

alansikora:
New docs for TXT records, also with a simple example. Please review.

@Sing-Li @MartinSchoeler @sampaiodiego

PR #1133, opened 2019-03-12
Update README.md

bpetrynski:
When connecting via default web SSH (on port 80), docker fails to start rocketchat because the port is already used: "Error starting userland proxy: listen tcp 0.0.0.0:80: bind: address already in use"


MartinSchoeler (approved):
Looks Good To Me! Thanks for the contribution!

PR #1134, opened 2019-03-13
Fix Documentation: Installation/Docker containers // Changed top command to lowercase

wopfel:
I guess this was by accident. At least I don't know a TOP uppercase command.


MartinSchoeler (approved):
Right! It must've been a typo. Thanks for the Contribution!

PR #1138, opened 2019-03-15
Install gnupg as a dependency (Debian 9)

VonUniGE:
Install gnupg as a dependency (Debian 9)

If gnupg is not installed the command will fail.


frdmn (approved):

frdmn:
Can confirm, when the `gnupg` package is missing the following error shows up: 

```
vagrant@stretch:~$ sudo apt-get install dirmngr && sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 2930ADAE8CAF5059EE73BB4B58712A2291FA4AD5
Reading package lists... Done
Building dependency tree
Reading state information... Done
dirmngr is already the newest version (2.1.18-8~deb9u4).
0 upgraded, 0 newly installed, 0 to remove and 19 not upgraded.
E: gnupg, gnupg2 and gnupg1 do not seem to be installed, but one of them is required for this operation
```

PR #1139, opened 2019-03-15
Update Documentation: Integrations

srpatel:
Explain how to send messages to other channels from a webhook. It's implied you can do this in the opening paragraph, but it's not explained how.


MartinSchoeler (approved):
Looks good to me! Thanks for the contribution!

PR #1173, opened 2019-03-26
Small correction on docker-compose installation step

juliomac:
Difficult to see this line on original documentation.

This is intended to improve it.

used this https://github.com/docker/compose/releases as reference.


grahamsmith:
Thanks for your contribution! This PR has gone stale and unfortunately fallen behind. We took the inspiration from this PR and have updated as such. Thanks again.

PR #1184, opened 2019-03-29
fixes #1183 overflow of table content due to table width

### LINKED ISSUE #1183: [UI] table content data overflow
Screenshot from 2019-03-29 23-44-28

The table content is overflow because the width is not large enough to hold it!

Screenshot from 2019-03-29 23-45-20

Setting the width of wrapper class to 100% will surely fix it I guess.

sudipt1999:
I increased the width of the table so that table content do not overflow the table

### DIFF
--- a/_sass/layout.scss
+++ b/_sass/layout.scss
@@ -436,7 +436,7 @@ main.home {
 
   .wrapper {
     padding-right: 1em;
-    max-width: 70%;
+    max-width: 100%;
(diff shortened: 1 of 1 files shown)


grahamsmith:
This probably _was_ useful but is no longer since moving to GB.

MartinSchoeler:
Agreed! Closing this issue. Thanks!

PR #1197, opened 2019-03-31
[Improvement] Make APIs example code blocks editable

shiqimei:
Hi, @MartinSchoeler ! I added a feature for the documentation. With this patch, users can replace the specific content before copying the example code as shown below.
demo

### DIFF
--- a/main.js
+++ b/main.js
@@ -284,6 +284,29 @@
     }
   }
 
+  // make code blocks editable
+  (function () {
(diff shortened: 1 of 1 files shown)


shiqimei (commented):
My implementation seems not so elegant and not so consistent with the original code base. Please review it and give me some advice. Thanks in advance 😘

PR #1203, opened 2019-04-02
[Feature] Add husky for auto-mdl before the push

shiqimei:
Hi, I found that our main repo uses husky as git hooks helper (See [here](https://github.com/RocketChat/Rocket.Chat/blob/64038492290b5523b5a674efe004c8ae3a064995/package.json#L237)). I try to transplant this feature to this repo so that we don't need to use `bundle exec "mdl ./ -r ~MD004 -i -g"` manually! The husky will run the above command automatically before every commit 😀

DeepinScreenshot_select-area_20190402220906

Please review my changes, thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .jekyll-metadata
 _site
 vendor/**
+**/node_modules/**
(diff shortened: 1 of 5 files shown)


sampaiodiego:
thanks for the contribution @lolimay .. the main repo uses `husky` mainly because it's a nodejs project, but this repo is not.

if you find a solution without having to install additional dependencies would be better.

MartinSchoeler (changes requested):
We should try to find an alternative in ruby

shiqimei:
> We should try to find an alternative in ruby

OK, I will try to implement this feature with Ruby packages.

PR #1213, opened 2019-04-08
Update the example result of REST API /api/v1/roles.list

shiqimei:


shiqimei:
The commit message `Update README.md` seems too ambiguous. Will close it and open another one :).

PR #1221, opened 2019-04-09
switch aws docker install from upstart to systemd

geekgonecrazy:


geekgonecrazy:
My biggest doubt is the systemd unit names.  Have we established any sort of standard naming?

MartinSchoeler:
@geekgonecrazy Can you fix the conflicts? Thanks!

PR #1222, opened 2019-04-09
fix docs links

cyberb:


MartinSchoeler (approved):
Looks good to me! Thanks for the contribution.

PR #1259, opened 2019-04-24
Add JFrog Xray integration

crazy-max:


crazy-max:
@MartinSchoeler Thoughts about this? Thanks :)

PR #1265, opened 2019-04-26
Update documentation

mastappl:
Update information about deployment via D2C. Fix links to images.


MartinSchoeler (approved):
Thanks for your contribution!

mastappl:
Hello @MartinSchoeler,

I've noticed the [note ](https://rocket.chat/docs/installation/paas-deployments/d2c-io/) about using with new RocketChat version.

The new stack is already in [our stackhub](https://d2c.io/stackhub/rocketchat-replicaset). Its' a separate one because of using ReplicaSet configuration as required. Should I update links in the current deployment page with D2C or create a new one (and leave the old one for using v.0)?

Regards,

PR #1266, opened 2019-04-26
Add note about mime-type for room.upload

geekgonecrazy:
Ref: https://github.com/RocketChat/Rocket.Chat/issues/14267

### DIFF
--- a/developer-guides/rest-api/rooms/upload/README.md
+++ b/developer-guides/rest-api/rooms/upload/README.md
@@ -26,7 +26,7 @@ curl "http://localhost:3000/api/v1/rooms.upload/GENERAL" \
     -H "X-User-Id: hw5DThnhQmxDWnavu"
 ```
 
-**Note: For some file types if uploading via curl you may need to set the mime type.**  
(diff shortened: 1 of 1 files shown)


wreiske:
I fully support and approve of this pull-request.

developer-guides/rest-api/rooms/upload/README.md
```diff
@@ -26,6 +26,10 @@ curl "http://localhost:3000/api/v1/rooms.upload/GENERAL" \
     -H "X-User-Id: hw5DThnhQmxDWnavu"
 ```
 
+**Note: For some file types if uploading via curl you may need to set the mime type.**
+
+With some file types it will set to `application/octet-stream`. You can can pass like this: `-F "file=@file.wav;type=audio/wav"`
```
wreiske, line 31: Dang, how'd we miss this? You can can... ahh!

geekgonecrazy: Wow... 🤦‍♂️ Committed to master and fixed.  Sheesh I need to practice my English 😂

PR #1269, opened 2019-04-27
check if request.content.message.sticker exists

### LINKED ISSUE #1270: Telegram integration example code fails, if message does not contain sticker
if a message has no text and no sticker, ln 17 will fail

arminfelder:
check if request.content.message.sticker exists, to prevent type error
Fixes #1270


MartinSchoeler (approved):
Thanks for the contribution

PR #1297, opened 2019-05-14
Add required step in instruction

robinreinhardt:
Add note to add @rocket.cat user in room


MartinSchoeler (approved):
Thanks for your contribution!

PR #1300, opened 2019-05-15
Update README.md

RSchlaeger:
The RocketChat server seems to have a problem with upper-case letters in the Hash digest. Therefore I propose to let users know this as well!


MartinSchoeler (approved):
Thanks for your contribution!

PR #1303, opened 2019-05-17
Update README.md

Alzavio:
Oplog is a paid feature on Heroku at a minimum of  usd18/month. It's no longer free nor very cheap.


MartinSchoeler (approved):
Thanks for your contribution!

PR #1305, opened 2019-05-21
Update the docker part

MyNameIsOka:
With this procedure, the necessary OPLOG / REPLICASET is also installed.
Startup scripts are not necessary anymore.


MartinSchoeler:
Hey @MyNameIsOka it seems that this PR is very outdated and might not fit the current doc. Let me know if you still think changes should be made to the current doc. Thanks!

PR #1306, opened 2019-05-21
Jira cloud and connecting bots to groups/channels

jdobbin-hsahealthplan:


d-gubert (changes requested):
Thanks for the contribution, @jdobbin-hsahealthplan !

I just think there's no need for the "*" 😄

PR #1316, opened 2019-05-24
Updating GCP deployment with Mongodb Replica set configurations

sshniro:


Sing-Li (approved):
Awesome.  Thanks for the tested, well written, update!

PR #1318, opened 2019-05-27
Fix stream name in example for stream-notify-room-users

dfaure-kdab:


MartinSchoeler (approved):
Thanks for your contribution!

PR #1325, opened 2019-06-01
New Documentation: Custom Sounds

pradumangoyal:
Document the Custom Sounds feature under the administrator tab

- Add a custom sound
- Using custom sound
- Delete/Edit custom sound

References #555


administrator-guides/custom-sounds/README.md
```diff
@@ -0,0 +1,27 @@
+# Custom Sounds
+
+Rocket.Chat allows you to add cutom sound that can be used as notification sound by other users for different channels.
```
MartinSchoeler: Maybe `custom sound files that can be used as the notification's sounds. Users can customize their notification sound preferences for specific channels`

administrator-guides/custom-sounds/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+You can add your own mp3 files.
```
MartinSchoeler: Maybe `Sound files should be uploaded in the mp3 format`

administrator-guides/custom-sounds/README.md
```diff
@@ -0,0 +1,27 @@
...
+- Go to the Administration tab
+- Select the Custom Sounds
+- Click on the Add symbol (+) in the right side bar
+- Enter a name for the sound
+- Browse and choose a mp3 file
+- Press the SAVE button
+
+## Using custom sound
+
+- To use the added sound as a notification for an channel select `Notifications Preferences` from the `More` section of the channel.
```
MartinSchoeler: Ideally, we should have this part on a dedicated `notifications preferences` article under the user guides and only link to said article. We should not "hide" user guides under an admin article

MartinSchoeler (changes requested):
Looks good, just need some minor adjustments

administrator-guides/custom-sounds/README.md
pradumangoyal: Yes, I thought to provide a link to the `Notification Preferences` here. But I was unable to find any `Notification Preference` article or any suitable heading under which it can come. So should I add `Notification Preference` somewhere and update this PR with the link?

MartinSchoeler (approved):
Looks good now! Thanks for your contribution!

PR #1354, opened 2019-06-26
Update README.md

reetp:
There is a link to a reverse proxy setup, but nothing to a multi proxy example.
This is just band aid. I think all the nginx configs should probably go in one place and then we just refer to 'Nginx' Setups or 'Apache' Setups


MartinSchoeler (approved):

behrouzamiri (commented):
the link in "Here is a link to a [Nginx multi instance reverse proxy example]" part is broken

PR #1357, opened 2019-07-04
Update installation manual

makhoff:
Added lines about installation to /usr/bin directory instead of /usr/local/bin.
The command "docker-compose --version" retrieves an error when installation is placed to /usr/local/bin


frdmn:
@makhoff Thanks for reporting this. Can you tell me which distribution you use? Is that a minimal one by chance? Usually `/usr/local/bin/` is part of the `$PATH` variable so placing the binary in there, should work fine for most distros.

I think we should stick to the regular location but add a separate note for people who only have `/usr/bin/` in their `$PATH`, just [like in the official Docker docs as well](https://docs.docker.com/compose/install/#install-compose-on-linux-systems).

frdmn:
Added a reference here: https://docs.rocket.chat/installation/docker-containers#3-installing-docker-and-docker-compose

PR #1358, opened 2019-07-05
Correct Rate Limit reset documentation

nemec:
I've been working on a project using the Rocket Chat API and have noticed a couple of places where the documentation is incorrect or out of date. First, the three rate limit properties seem to be sent with _every_ API call (tested on `/api/v1/channels.history` at least) and second the rate limit reset header seems to be provided in _milli_seconds, not seconds. Example from a server I'm on: `1562287945706`.

Are these behaviors configurable by the server owner (in which case I assume additional doc changes are necessary)?

### DIFF
--- a/developer-guides/rest-api/rate-limiter/README.md
+++ b/developer-guides/rest-api/rate-limiter/README.md
@@ -10,9 +10,9 @@ or by providing a valid configuration object with the desired configuration, in 
 
 The rate limiter is set by default in the development environment and can be disabled in the admin panel in the `Administration Panel => Rate Limiter => API Rate Limiter`.
 
-For requests that are made and **exceed** the limit set by the rate limiter, three extra properties will be sent by the request headers.
+For requests that are made to an api endpoint with a rate limiter, three extra properties will be sent by the request headers.
 
 1. `x-ratelimit-limit`: The number of calls allowed for a certain amount of time.
 2. `x-ratelimit-remaining`: The number of requests remaining in the current rate limit window.
-3. `x-ratelimit-reset`: The time at which the current rate limit window resets in [UTC epoch seconds](https://en.wikipedia.org/wiki/Unix_time).
+3. `x-ratelimit-reset`: The time at which the current rate limit window resets in [UTC epoch milliseconds](https://en.wikipedia.org/wiki/Unix_time).
 


developer-guides/rest-api/rate-limiter/README.md
```diff
@@ -10,9 +10,9 @@ or by providing a valid configuration object with the desired configuration, in
 
 The rate limiter is set by default in the development environment and can be disabled in the admin panel in the `Administration Panel => Rate Limiter => API Rate Limiter`.
 
-For requests that are made and **exceed** the limit set by the rate limiter, three extra properties will be sent by the request headers.
+For requests that are made to an api endpoint with a rate limiter, three extra properties will be sent by the request headers.
```
MarcosSpessatto, line 13: Good catch @nemec ! Thank you!

developer-guides/rest-api/rate-limiter/README.md
```diff
@@ -10,9 +10,9 @@ or by providing a valid configuration object with the desired configuration, in
...
 
 1. `x-ratelimit-limit`: The number of calls allowed for a certain amount of time.
 2. `x-ratelimit-remaining`: The number of requests remaining in the current rate limit window.
-3. `x-ratelimit-reset`: The time at which the current rate limit window resets in [UTC epoch seconds](https://en.wikipedia.org/wiki/Unix_time).
+3. `x-ratelimit-reset`: The time at which the current rate limit window resets in [UTC epoch milliseconds](https://en.wikipedia.org/wiki/Unix_time).
```
MarcosSpessatto, line 17: Yes, it's seconds. According to the Wikipedia description: 

> Unix time (also known as POSIX time[1][2] or UNIX Epoch time[3]) is a system for describing a point in time. **It is the number of seconds** that have elapsed since 00:00:00 Thursday, 1 January 1970,[2] Coordinated Universal Time (UTC), minus leap seconds.

MarcosSpessatto (changes requested):

developer-guides/rest-api/rate-limiter/README.md
nemec: I know the definition of UNIX time is seconds, but the Rocket API on the server I tested returns the number of _milliseconds_ since Jan 1, 1970 in the x-ratelimit-reset header. If I take the number the API gave me (`1562287945706`) as _seconds_, it ends up being in the year 51476 instead of 2019.

I just wanted to make sure the docs reflected the actual value that is being sent back

MartinSchoeler: Ideally, I think we should change the code to work with seconds to reflect the UNIX Epoch time standard. But for now, I think we should merge this.

MartinSchoeler (approved):

MartinSchoeler:
Thanks for your contribution!

PR #1361, opened 2019-07-11
Update README.md

marco-aa:
as discussed in the support channel - result from my question - nice help - thanks


marco-aa:
found a mistake - forgotten link - will make new PR

PR #1370, opened 2019-07-30
added a note on important env variable

idarlund:
Added a note so that people configuring reverse proxy in front of rocket.chat remembers to set correct ```HTTP_FORWARDED_COUNT``` for correct clientAddress calculation


geekgonecrazy (approved):

geekgonecrazy:
Thanks for adding this!

PR #1372, opened 2019-08-03
changed to correct header setting

idarlund:
HTTP header settings are wrong.

It should be ```X-Forwarded-For``` and not ```X-Forward-For```. Have look at https://en.wikipedia.org/wiki/X-Forwarded-For.

Updated the documentation accordingly.


MartinSchoeler (approved):
Looks good to me

PR #1374, opened 2019-08-05
Update docker-compose file example in README.md

cb3inco:
The docker-compose example had rocket.chat:latest for the rocket.chat image, however, a more accurate one is: rocketchat/rocket.chat:latest which points to the correct docker hub image path.


johncrisp:
Hi, 

Is it possible to comment/add a warning to the file to suggest that :latest may cause issues, or that setting a fixed version is a better practice?

We have had a number of incidents where an older version has been tagged as latest which has caused issues for users.

PR #1388, opened 2019-08-16
Manage user with only the first name filled.

dungtri:
Hello,

I found many users in telegram which has only fill the first name as minimal info.
In that case 'undefined' word are displayed in rocket chat user/channel.

> Dung Tri Undefined say.. blablabla

I propose a fix for that.

A pleasure to contribute to this very great project !

Dung Tri.


MartinSchoeler (approved):
Thanks for your contribution!

PR #1393, opened 2019-08-20
Fix path

hever:


geekgonecrazy:
@LuluGO I think you are on debian.   Can you confirm this is right?

hever:
Few line before is

> `cat << EOF |sudo tee -a /lib/systemd/system/rocketchat.service`

and then

> Open the Rocket.Chat service file just created (`/usr/lib/systemd/system/rocketchat.service`)

Faria1212 (approved):

Faria1212 (approved):

geekgonecrazy (approved):

geekgonecrazy:
Looks like some how this change was made outside of this PR.  :(

PR #1409, opened 2019-09-18
customFields url support explained compare PR #15403

marco-aa:
Related to PR https://github.com/RocketChat/Rocket.Chat/pull/15403

Explains that customFields of type text now can contain a url wich will be rendered in the usertab


Faria1212:
Merged manually. Thank you!

PR #1463, opened 2019-10-14
Add doc about direct message import via CSV

rodrigok:


MartinSchoeler (approved):

felipeozalmeida:
Hello guys!

First of all, thank you for your awesome work.

I don't know if this is the proper place, but I've found possibly a mispelled folder name in that PR, that is, "directmessage" or "directmessages". These two names appear in the same page and I've found them when reading the docs to learn how to import things via CSV and got confused.

I could fix that if I knew the right name, but I'm not sure.

Thank you guys!

rodrigok:
@felipeozalmeida thanks, here is the fix: https://github.com/RocketChat/docs/pull/1468

PR #1495, opened 2019-10-29
Fixed Typo

zdravst:
tranlsated TO translated


Rodriq:
Greetings @zdravst I suggest it's better you combine the PRs #1495 #1494 #1493 #1491 #1492 as one so it'll be easy to merge.
  They're both fixing typos and were open at almost the same hour.
What do you think @MarcosSpessatto ?

MartinSchoeler (approved):

PR #1501, opened 2019-10-31
Fix Documentation: Deploying on AWS

georgeos:
Fix on the service name for mongo


MartinSchoeler:
Hey @georgeos it seems that this PR is very outdated and might not fit the current doc. Let me know if you still think changes should be made to the current doc. Thanks!

PR #1505, opened 2019-11-05
Create External API Calls

paulofariadias:


security/External API Calls
```diff
@@ -0,0 +1,86 @@
+# External API Calls
```
MartinSchoeler, line 1: Please add newlines before and after each title (on the first title only after)

security/External API Calls
```diff
@@ -0,0 +1,86 @@
...
+Rocket.Chat comes with a range of features that call APIs for different external services. This provides more information about the data sent by Rocket.Chat services in API calls to external services.
+## Any URL
```
MartinSchoeler, line 5:
```suggestion

## Any URL

```
Like this

MartinSchoeler (changes requested):
Asked for some changes, only to satisfy the linter

MartinSchoeler (approved):

PR #1529, opened 2019-12-06
Update to fit new password setting rules

makibras:
password can no longer be manually set, so it must be set via email, as any other user


makibras:
regression

PR #1560, opened 2020-01-07
Fixed versions of Docker Compose and Mongo Images

lucasduete:
...to solve problems on AWS installation guide


grahamsmith:
Thanks for your contribution! Sadly this PR has gone stale and the docs have been updated since this PR was opened, as a result we are going to close it. Thanks agin!

PR #1570, opened 2020-01-26
Update README.md

KnifeParty12:
Fixed typo in testing documentation.


Fishbowler:
Unsure if this makes it clearer.
How about "This will set the duration of all animations to 0"?

MartinSchoeler:
This is not a typo, please check [Rule 2a](https://www.grammarbook.com/punctuation/apostro.asp). Thanks!

PR #1580, opened 2020-02-04
Add Mobile Apps Accessibility idea - GSoC 2020

cardoso:


Sing-Li (approved):
Very high demand, vital topic.    Might have to be done in collaboration with web client accessibility though.   Would love to see some new ideas from the student and industry.

PR #1581, opened 2020-02-04
[GSoC 2020] Add Omnichannel project ideas

renatobecker:


Sing-Li (approved):
Awesome list of ideas!     Rasa I think will be "hot".    We might need to do some "training" and evangelism before proposal time with the non-open-source expertise required ... especially in Zoho and Dialogflow.

PR #1616, opened 2020-02-21
Update projects.yml

paulofariadias:
Added Message Auditing


MartinSchoeler:
No need for adding this article to the home page for now

PR #1643, opened 2020-03-12
Docker Swarm Installation

VirtualEstatePlanner:
README, yml template, and convenience script


VirtualEstatePlanner:
I'd like to add docker secrets as well as gitlab, ldap, and nextcloud integrations into this example later.

MartinSchoeler (changes requested):
Could you add the new files to the [table of contents](https://rocket.chat/docs/contributing/documentation/contribution-guidelines/#table-of-contents)?

Thanks!

VirtualEstatePlanner (commented):
I believe this is the line you wanted me to add.

VirtualEstatePlanner (commented):
I'm a bit incredulous that I even have the option to verify that I viewed the very same files that I wrote and committed, but I definitely did also view these files.

VirtualEstatePlanner (commented):
So, I viewed my own edits again.  This feature is weird to me.  Can a brother get a code review?

installation/docker-containers/docker-swarm/README.md
```diff
@@ -0,0 +1,22 @@
+This is the easiest way to setup RocketChat on Docker Swarm:
+
+  1. Make sure your Cloudflare dashboard shows correct DNS entries for `traefik.${DOMAIN_WITH_DNS_HANDLED_BY_CLOUDFLARE}` and `rocketchat.${DOMAIN_WITH_DNS_HANDLED_BY_CLOUDFLARE}` pointing to your docker swarm.
```
geekgonecrazy: This should be tweaked.  We should not be assuming that someone is using cloudflare for dns.  Using CLOUDFLARE through out the whole document is just going to lead to confusion

geekgonecrazy:
I like the idea of this.  But i'm kinda bothered that this basically is hard coded through out everything assuming the person is using cloudflare.

While I have nothing against cloudflare.. I don't think we should be forcing cloudflare in any of the documentation.  In other documentation we try to leave DNS handler completely up to the person setting it up.

Would it be possible to adjust to be a bit more flexible on the dns provider?  Maybe replace references to something more generic like DOMAIN?

This sort of instruction as is would be 100% perfect for a blog or tutorial where opinionated is acceptable.  But here being opinionated towards cloudflare as DNS I think is not something most people would take well.

VirtualEstatePlanner:
> I like the idea of this. But i'm kinda bothered that this basically is hard coded through out everything assuming the person is using cloudflare.
> 
> While I have nothing against cloudflare.. I don't think we should be forcing cloudflare in any of the documentation. In other documentation we try to leave DNS handler completely up to the person setting it up.
> 
> Would it be possible to adjust to be a bit more flexible on the dns provider? Maybe replace references to something more generic like DOMAIN?
> 
> This sort of instruction as is would be 100% perfect for a blog or tutorial where opinionated is acceptable. But here being opinionated towards cloudflare as DNS I think is not something most people would take well.

I chose cloudflare because it is possible to use freely (as in beer).  There are several other options available for traefik to pull Let's Encrypt certs, but I personally can't afford to purchase paid services on all of them and then test and write configs for them.  If someone has another DNS provider, they are free to modify the example code based on traefik's documentation of their provider.  Of course, the user can also just comment out the traefik section and use nginx or HAProxy as a reverse proxy if they so choose.

Edit - clarity: traefik was chosen because it handles microservice environments very well, and Let's Encrypt requires DNS validation to generate wildcard certificates, as this example does (which is also very useful in a microservice environment).

geekgonecrazy:
No problem with traefik at all.  We use traefik heavily and gladly support it.  I'd recommend swapping over to `HTTP-01` for letsencrypt that way you do not need to talk to a specific dns provider and then the instructions become agnostic again to the dns provider which is how we would prefer it.

The idea is I might be running route 53 or digitalocean or {insert dns provider here} when I come try to install Rocket.Chat I should not need to worry about moving to a dns provider just to be able to follow a guide.

Like I said 100% perfect for a blog post or article but we need it a tad bit more flexible to be in our documentation.  That way its useful for the most amount of people.  

Think of it imagining you were the one that got contacted everytime someone tried to follow this and their existing dns was at their domain registrar with tons of subdomains defined already and not cloudflare.  This is a key thing for us to keep in mind on our documentation otherwise we are the ones that get those support requests :grin:

VirtualEstatePlanner:
> No problem with traefik at all. We use traefik heavily and gladly support it. I'd recommend swapping over to `HTTP-01` for letsencrypt that way you do not need to talk to a specific dns provider and then the instructions become agnostic again to the dns provider which is how we would prefer it.
> 
> The idea is I might be running route 53 or digitalocean or {insert dns provider here} when I come try to install Rocket.Chat I should not need to worry about moving to a dns provider just to be able to follow a guide.
> 
> Like I said 100% perfect for a blog post or article but we need it a tad bit more flexible to be in our documentation. That way its useful for the most amount of people.
> 
> Think of it imagining you were the one that got contacted everytime someone tried to follow this and their existing dns was at their domain registrar with tons of subdomains defined already and not cloudflare. This is a key thing for us to keep in mind on our documentation otherwise we are the ones that get those support requests 😁

I would think the ideal solution there would be to show a handful of different options (I'd say the most common DNS providers you guys encounter - probably something like Route53, Google DNS, DigitalOcean, Cloudflare, whatever Azure's DNS provider is called, etc.) in separate subfolders of `installation/docker-containers/docker-swarm`, and include a link to both traefik's and Let's Encrypt's respective documentation pages regarding the `DNS-01` challenge.  That should theoretically minimize to the greatest degree the total number of support calls you receive while also minimizing the effort to install the application on Docker Swarm for the greatest total number of users.

While I do not personally have access to these platforms for testing, it is likely that modifying the existing Cloudflare example for those platforms should be very easy, perhaps even trivial for your team if you have access to those platforms.  I am willing to help in whatever way I can.

engelgabriel (approved):

PR #1676, opened 2020-04-08
Fixes when you use federation on a subdomain

pierreozoux:
Currently the documentation is misleading if you use it on a subdomain.
This proposes to fix that.


Sing-Li:
@pierreozoux awesome.  Thanks for the doc-fix :slightly_smiling_face:   please add to this PR as you hit more brick walls  (like only private groups can be federated ...).    

Will check again once you've successfully federated for the final merge.

Sing-Li (approved):

PR #1679, opened 2020-04-08
Update README.md (rest-api/users/setstatus)

schmik:
https://rocket.chat/docs/developer-guides/realtime-api/method-calls/user-presence/ list possible values as `online`, `away`, `busy`, `offline` -- **`offline`** , not `invisible`. 

POST with '{"message":"My status update", "status": "invisible"}' will cause 'BAD REQUEST', while POST with `offline` will succeed.


MartinSchoeler (approved):
Thanks for the contribution!

PR #1683, opened 2020-04-11
Fix Documentation: Rocket.Chat in Ubuntu

jordan808:
There is a missing "sudo" in the command `cd /tmp/bundle/programs/server && npm install` . This creates a few errors in downloading dependencies that need sudo privileges.

### DIFF
--- a/installation/manual-installation/ubuntu/README.md
+++ b/installation/manual-installation/ubuntu/README.md
@@ -72,7 +72,7 @@ tar -xzf /tmp/rocket.chat.tgz -C /tmp
 Install (this guide uses /opt but feel free to choose a different directory):
 
 ```bash
-cd /tmp/bundle/programs/server && npm install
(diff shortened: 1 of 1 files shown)


installation/manual-installation/ubuntu/README.md
```diff
@@ -72,7 +72,7 @@ tar -xzf /tmp/rocket.chat.tgz -C /tmp
 Install (this guide uses /opt but feel free to choose a different directory):
 
 ```bash
-cd /tmp/bundle/programs/server && npm install
+cd /tmp/bundle/programs/server && sudo npm install
```
frdmn, line 75: There should be no `sudo` necessary, since the modules will be installed into the current working directory - which then will be changed to `rocketchat:rocketchat` later anyways.

frdmn (changes requested):

PR #1693, opened 2020-04-18
[NEW] Add the ability to share Account custom fields to external applications

renatobecker:
We are now supporting a new custom field property for user accounts, called `sendToIntegrations`.
The new property will allow you to share the fields that have this property with external systems, such as CRM, external applications and so on.

The Omnichannel CRM integration will use this new feature, sharing all of the user custom fields that have the `sendToIntegrations` configured, like the example below:

```
{
...
    sendToIntegrations: true,
...
}
```

To get more details about this new feature, please take a look at [this PR](https://github.com/RocketChat/Rocket.Chat/pull/16286).

### DIFF
--- a/administrator-guides/custom-fields/README.md
+++ b/administrator-guides/custom-fields/README.md
@@ -29,11 +29,17 @@ Make sure to use a valid `JSON`, where `keys` are the `field names` containing a
   "required": true,
   "minLength": 2,
   "maxLength": 10
+ },
+ "crmUserId": {
+  "type": "text",
+  "required": false,
+  "minLength": 12,
+  "sendToIntegrations": true,
  }
 }
 ```
 
-In the example above we set two new fields with the following properties:
+In the example above we set three new fields with the following properties:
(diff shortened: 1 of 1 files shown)


administrator-guides/custom-fields/README.md
```diff
@@ -57,4 +63,6 @@ In the example above we set two new fields with the following properties:
 
 - **private**: defines the field as private, so only users with `view-full-other-user-info` permission can see this field when viewing this user's profile.
 
+- **sendToIntegrations**: define the field as shareable with external applications, such as Omnichannel integrations.
```
sampaiodiego, line 66: At some point I think we'll need to describe better which integrations the field will be sent to.. but for now I think Omnichannel integrations is fine 👍

sampaiodiego (approved):

renatobecker:
@marceloschmidt can you please merge this PR?
I don't have permission to do it, thanks.

PR #1695, opened 2020-04-22
Update Cloudron.io demo link

gramakri:
We recently changed the location of our demo instance to my.demo.cloudron.io


MartinSchoeler (approved):
Thanks for the update on our docs!

PR #1699, opened 2020-04-24
 How to resume login session with authtoken

premdasvm:
added how to resume login session with authtoken. current it's only found in SDK


MarcosSpessatto (changes requested):
@premdasvm Actually we don't support the resume as you can see here: https://github.com/RocketChat/Rocket.Chat/blob/develop/app/api/server/api.js#L408

frdmn:
Will close this PR then.

PR #1711, opened 2020-05-09
Supported OS : ubuntu 20.04 added

iamayushdas:
Rocket.chat is running IN 20.04LTS as same as it was running in 18.04LTS.


MartinSchoeler (approved):
LGTM Thanks for the contrib!

PR #1715, opened 2020-05-12
Hyper.sh went out of business in early 2019

fbartels:
https://news.ycombinator.com/item?id=18734658

Signed-off-by: Felix Bartels <felix@host-consultants.de>


MartinSchoeler (approved):
Makes sense to remove the docs then. Thanks!

PR #1726, opened 2020-05-19
Fix build

Fishbowler:
Make htmlproofer happy by fixing broken links


Fishbowler:
Hmmmm. Just fixing the html-proofer doesn't fix the build. Gotta fix the linting too. On it..

SUMMARY.md
```diff
@@ -2,563 +2,563 @@
```
MartinSchoeler: This file is generated by gitbook, so I think any changes done here will be reverted by it

MartinSchoeler (changes requested):
I think only one file needs to be changed

SUMMARY.md
Fishbowler: I've modified the commit and removed this file, rebased, force pushed.

Fishbowler:
I think this is all fixed up?

MartinSchoeler (approved):
Looks good to me! Thanks for helping out!

PR #1737, opened 2020-05-29
Show actual backticks in help for inline code

marcjansen:
Before this the help wouldn't explain how to achieve inline code:

screenie

An alternative would be to actually use a `<code>`-block:

```
* Inline code: <code>`Lorem ipsum dolor`</code>
```

Please review.


grahamsmith (approved):

grahamsmith:
@marcjansen great spot :)

marcjansen:
> @marcjansen great spot :)

Thanks. Unsure why the build failed, though.

grahamsmith:
At the moment the docs are moving to GitBook and the tooling around the CI uses Jekyll and Liquid however Git Book adds none standard tags and MD so the CI fails. Work is active to solve this I am told so its nothing you have done.

engelgabriel (approved):

PR #1740, opened 2020-06-03
Fix Documentation: Removed redundant image with architecture diagram

bandantonio:


grahamsmith (approved):
Thanks for spotting and fixing this 👍

PR #1754, opened 2020-06-09
Hugodecasta update ssl config

hugodecasta:
Specified the Intermediate Certificate file path directly in the v-host conf file using the `SSLCertificateChainFile`.


installation/manual-installation/configuring-ssl-reverse-proxy.md
```diff
@@ -110,8 +110,6 @@ Lock down permissions: `chmod 400 /etc/ssl/private/chat.domain.com.key`
 
 Add your certificate to `/etc/ssl/certs/chat.domain.com.crt`
 
-Add your intermediate to `/etc/ssl/certs/intermediate.ca.pem`
```
frdmn: Shouldn't we keep this instruction (to create the `intermediate.ca.pem` file) since it's still referened in the `SSLCertificateChainFile` directive?

hugodecasta: Yes actually, i deleted that from an older config.
I confirm we need to keep this one !

frdmn (approved):

frdmn:
Thanks for your contribution, Hugo!

PR #1758, opened 2020-06-15
IBM Bluemix is gone...

jjasghar:
- IBM Bluemix is no-longer used. Everything is IBM Cloud now


MartinSchoeler (approved):
Indeed, thanks for the contribution!

PR #1759, opened 2020-06-22
Added 'ssl http2' to the nginx vHost

DecaTec:
Without the `ssl`, no SSL connection can be established (e.g. Firefox says "SSL_ERROR_RX_RECORD_TOO_LONG"). So this one is mandatory.

`http2` is for HTTP/2 support which should be save to use.


DecaTec:
E

DecaTec:
The SSL config is alredy enabled by `ssl on;`

PR #1776, opened 2020-08-13
[FIX] change query parameter for livechat history

antkaz:
Parameter `ts` changed to `ls`


MartinSchoeler (approved):
It checks out with the code

Issue #1796, opened 2020-10-01
Missing "install" keyword in command to install required tools


aradhya-gupta:
Under the "Developing in Windows 10"  section of the Developer Guides, the command to install required tools is incomplete and does not work due to the missing "install" keyword.
It should be "sudo apt-get **install** build-essential git curl python-minimal" instead of "sudo apt-get build-essential git curl python-minimal"

PR #1801, opened 2020-10-08
Updated command for Install tools required

RohitKumar-200:
Changed "sudo apt-get build-essential git curl python-minimal" to "sudo apt-get install build-essential git curl python2-minimal" . Added install keyword and changed python to python2, bacause python-minimal package is replaced by python2-minimal.
[Fixed] error "E: Invalid operation build-essential"


RohitKumar-200:
#1796

Faria1212 (approved):

PR #1803, opened 2020-10-12
Add missing install command in instructions

JaredHonda:
I found this problem while trying to run the command for installing the required tools via WSL 2 shell.

The 'install' keyword is missing, which causes the installation process to fail. This minor text edit includes this keyword so that it doesn't block users during the installation process while they follow along with the guide.


MartinSchoeler:
Duplicate of https://github.com/RocketChat/docs/pull/1801/

PR #1815, opened 2020-10-22
mongod and nginx are not targets

ghost:
I was wondering why my rocketchat was starting and erroring before mongodb started.


ghost:
Also I don't think there's any reason to wait for nginx to start?

Sing-Li (approved):

PR #1816, opened 2020-10-22
Fix codeblock examples and a typo

tuelwer:
The code block needs to be indented. Otherwise the backticks will not be displayed correctly.


Sing-Li (changes requested):
Please revert the code blocks mod as it is incorrect.   And it has been fixed by another PR.

We will merge the typo fix after you remove the code block fix.      Please consider one fix per PR as a best practise.

PR #1826, opened 2020-11-15
Update manual install with needed NodeJS version

fallen:
Without this change, RocketChat does not start anymore with NodeJS 12.14.0.


teak421:
New enterprise customer here... I got caught by this.  Docs should be updated.  In all fairness though, I should have checked out the node version on the release page here.  Still, it would be nice to close the confusion.

frdmn (approved):

PR #1843, opened 2020-12-11
Fix field name for customFields in updateOwnBasicInfo

### LINKED ISSUE #1842: customFields are not updated with users.updateOwnBasicInfo
On https://github.com/RocketChat/docs/blob/master/api/rest-api/methods/users/updateownbasicinfo.md it is stated that we should send `data.customFields`, but it should be `customFields`, as seen on https://github.com/RocketChat/Rocket.Chat/blob/master/app/api/server/v1/users.js#L497

So the payload for updating customFields using `users.updateOwnBasicInfo` should be

```json
{
	"customFields": {
			"company": "Test"
		}
}
```

and not as currently stated
```json
{
	"data": {
		"customFields": {
				"company": "Test"
			}
	}
}
```

I'll open an issue for that but as PR seems to be merged slowly I'll leave this open for anyone having the same issue...

HugoHeneault:
Closes #1842


MartinSchoeler (approved):
Looks good to me!

PR #1844, opened 2020-12-14
Update developing-on-windows-10.md

wg-connect2ddentistry:
I've added a command "install" which is required with apt-get, and one dependency which is required during the npm install "pkg-config" 
experience on Ubuntu 18.04


rodrigok (approved):

rodrigok:
@wg-connect2ddentistry thanks for your contribution, our dev docs were moved to developer.rocket.chat and your suggestion was applied there.

PR #1855, opened 2021-01-07
remove heroku.md and corresponding mentions

aditya-mitra:
[This Pull Request **on the RocketChat Main Repository**](https://github.com/RocketChat/Rocket.Chat/pull/19901) had removed heroku from the README.

Heroku can no longer be used as a PAAS deployment method (unless used with docker).

The *heroku.md* states the use of **mLab MongoDB addon** which is discontinued. (https://devcenter.heroku.com/changelog-items/1823)

This was also [creating issues](https://github.com/RocketChat/Rocket.Chat/issues?q=is%3Aissue+is%3Aopen+heroku) that the deployment was failing because users are still using the *One-Click Heroku Deploy* just to find that it was not working. Heroku is a very popular deployment method.
 
Related issues: 
- https://github.com/RocketChat/Rocket.Chat/issues/19562
- https://github.com/RocketChat/Rocket.Chat/issues/18855
- https://github.com/RocketChat/Rocket.Chat/issues/18570

Also heroku should not be used a deployment method because:
- Heroku works like a cloud function. So, it does a cold start and cannot bind to port within 60 seconds. 
(Related Issue - https://github.com/RocketChat/Rocket.Chat/issues/8051)
This also makes RocketChat to take several minutes to start again. 
- Heroku only provides a **storage** of 500 MB in the free tier. In some cases (might be more as RocketChat grows), the size of the app exceeds this limit. Heroku stops the app as soon as the app crosses this limit.


MartinSchoeler (approved):
Indeed, a lot has changed with heroku and we should update our documentation to reflect that

PR #1856, opened 2021-01-07
Add a note on Rate limiter not always tweakable

HugoHeneault:
Some endpoints (like `/api/v1/users.updateOwnBasicInfo`) might trigger the rate limiter and cannot currently be customized/disabled in `Administration Panel` so I added a note about it.


Faria1212:
Merged manually! Thanks

PR #1865, opened 2021-01-25
It is not necessary to provide _id when posting to this endpoint.

JayBraker:
Maybe I got the desciption of how _id s of messages are derived wrong  but at least it is not required to provide an _id when posting to the endpoint.
Does the magic happen in the background?


MartinSchoeler (approved):
Indeed, by reading the code it seems that the Id is not explicitly required, only the message object param.

PR #1867, opened 2021-01-28
Update security-updates.md

Nono-m0le:
Wrong date / number


MartinSchoeler:
Thanks for your diligence. To avoid any doubt with years in the future, we removed the year from the numbering.

PR #1877, opened 2021-02-12
Fix gitlab wehook help

wesleyjellis:
The old link appears dead, I think https://docs.gitlab.com/ee/user/project/integrations/webhooks.html is the new correct link


MartinSchoeler:
@wesleyjellis are you sure that the provided link is correct? I can't seem to access it, how about https://docs.gitlab.com/ee/user/project/integrations/webhooks.html

wesleyjellis:
:facepalm: oops, good catch!

MartinSchoeler (approved):

MartinSchoeler:
Awesome, Thanks!

PR #1885, opened 2021-02-22
[TYPO] Change "ts" param of livechat/message to "ls"

yash-rajpal:
The livechat/message api expects the param `ls` for date as loading start but in docs its mentioned that it expects `ts` which creates confusion. 

Check the param requirement [here](https://github.com/RocketChat/Rocket.Chat/blob/eabb30472bd16a4000465ef1f06e5404dc4c1e2e/app/livechat/server/api/v1/message.js#L232)

This PR corrects this typo.

### DIFF
--- a/api/rest-api/methods/livechat/message.md
+++ b/api/rest-api/methods/livechat/message.md
@@ -47,7 +47,7 @@ curl -X POST \
       "username": "guest-4",
       "name": "Livechat Visitor"
     },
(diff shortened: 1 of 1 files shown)


yash-rajpal:
@renatobecker Please have a look, documentation for livechat/message api shows a wrong param. :)

MartinSchoeler (approved):
Indeed! I've checked the endpoint code and this PR is correct

PR #1886, opened 2021-02-25
Fixing Wrong command in WSL2 guide

yash-rajpal:
### Earlier
The command given to install packages on WSL2 is - `sudo apt-get build-essential git curl python-minimal` which is not a valid command and produces the error below:- 
image

### Now
We are missing `install` argument here, command should be -
`sudo apt-get install build-essential git curl python-minimal` which installs the mentioned packages.
image

Here we can see packages being installed.


yash-rajpal:
@Sing-Li Please have a look. Thanks :)

PR #1931, opened 2021-04-22
Fix typos in Create New team page

shrinish123:
There were some typos in the create new Team page. It should be 'click on' instead of 'click in'


shrinish123:
Please Review @Faria-TechWrite 😄

PR #1934, opened 2021-04-23
Fix Grammatical mistakes

shrinish123:
There were some small mistakes where only one gender was taken into consideration while addressing the user,I have changed it to neuter gender


shrinish123:
@Faria-TechWrite Please Review 😄

PR #1942, opened 2021-04-23
Correcting typos/adding consistency for the Administrator Guides in: Import/AccountsSettings/Federation

Jeanstaquet:
I have added consistency on each page regarding the synthaxe/code block/alt text that was sometimes missing. For example, often in a list with items, every other enumeration does not end with a full stop at the end of the sentence.


Faria1212:
Merged manually. Thanks!

PR #1947, opened 2021-04-24
Correcting typos/adding consistency in Mobile Guides & Security and Compliance Guides

Jeanstaquet:
I have added consistency on each page regarding the synthaxe/code block/alt text that was sometimes missing. For example, often in a list with items, every other enumeration does not end with a full stop at the end of the sentence. Or a code block was in 'text' form when it is Js...


Faria1212:
Action items/steps/instructions starting with numbered list should not end in a period. But thank you so much for the effort. https://handbook.rocket.chat/product/technical-writing

PR #1949, opened 2021-04-25
Fix Mistakes in Accounts Page

shrinish123:
There were places where sentences were constructed taking into consideration only one gender.


shrinish123:
Please Review @Faria-TechWrite 😄

Faria1212 (approved):

PR #1951, opened 2021-04-27
Fix passive voice overuse

Kartik18g:


Kartik18g:
@Faria-TechWrite can you review it :)

PR #1971, opened 2021-05-04
Update snap install instructions in README.md for network-observe plug

andweber:
Fix in https://github.com/RocketChat/Rocket.Chat/pull/20852 is not active by default. 

Additional see: https://github.com/RocketChat/Rocket.Chat/issues/18086

Should be activated manually (proposed here with an update in the docs) or alternativley there is a way of auto-activate - which is at least unknown to me.


engelgabriel:
@geekgonecrazy can. you review this?

geekgonecrazy:
@andweber not sure I understand the need for this to be recommended as attached.  It works with out doing this.  What does attaching the plug fix or improve?

andweber:
well, with current version 3.13.2 (on Ubuntu 20.04 LTS) which includes the fix https://github.com/RocketChat/Rocket.Chat/pull/20852 I still observe the issues mentioned in https://github.com/RocketChat/Rocket.Chat/issues/18086 and https://github.com/RocketChat/Rocket.Chat/issues/14562 (which are by the way the same). Thus the fix seems not yet to work correctly - at least for me and two other mentions in https://github.com/RocketChat/Rocket.Chat/issues/18086. 

In the documentation of snaps https://snapcraft.io/docs/network-observe-interface the "network observe" plug is mentioned as "auto-connect: no". I only have a very limited understanding of snap - but as far as I understood the documentation the plug must be manually connected to work. 
The plug should only work out of the box in "dev mode" and "classic confinement". Neither is recommended in the rocket chat snap install instruction.
 
Manually connecting the plug, solves the above issue in version 3.13.2. for me, that's why I assume this has to be done in general and recommended the change.

geekgonecrazy (approved):

andweber:
https://github.com/RocketChat/Rocket.Chat/issues/18086 can be closed as fixed.
(wasn't closed automatically?)

PR #1986, opened 2021-05-29
Update and rename bitbucket-server-integration.md to bitbucket-cloud-…

maikondouglas:
…integration.md

Request to update the Bitbucket Cloud installation/integration manual


d-gubert:
Hey @maikondouglas ! We'd be happy to have docs regarding our Bitbucket Cloud app, but the instructions to use it are different from the Bitbucket Server Integration app, which is a valid app btw, so we likely shouldn't remove those docs 🙂

d-gubert (changes requested):
Hey @maikondouglas ! We'd be happy to have docs regarding our Bitbucket Cloud app, but the instructions to use it are different from the Bitbucket Server Integration app. Also, the Bitbucket Server Integration is a valid app, so we likely shouldn't remove those docs 🙂

maikondouglas:
I think it's interesting so you're creating this documentation for Bitbucket Cloud @d-gubert

PR #1997, opened 2021-07-19
Update supported MongoDB versions

amottier:
As MongoDB version 3.6 reach end of life it seems to make sense to recommend usage of version 4.2 as it is the most recent version tested in CI (https://github.com/RocketChat/Rocket.Chat/blob/3.16.3/.github/workflows/build_and_test.yml#L196)


amottier:
This PR is related to RocketChat/feature-requests#241.

PR #2011, opened 2021-09-10
Fixed typo in restore command to stop the snap

simonbooth:


Faria1212:
Rectified already! Thank you

PR #2020, opened 2021-10-13
Removing broken screenshot URLs from imageforge

aksakalli:


debdutdeb (commented):
Thanks for the catch and fix. I think it'd be better to have a replacement screenshot/code block with the command output there since the follow up text points to such implication.

Rodriq (commented):
This is a valid change.

Rodriq:
Because of some conflicts issues. I am going to do the fix directly from Gitbook with a screenshot of the output of the command.
This will be closed now. Thanks!

PR #2022, opened 2021-10-29
improved docs for supported mongo versions

blacksheep557:


quick-start/installing-and-updating/manual-installation/mongo-versions.md
```diff
@@ -2,6 +2,8 @@
 
 Rocket.Chat currently requires MongoDB version 3.6 or higher. [Support of MongoDB 3.2 will has been removed from Rocket.Chat 2.0.0 and MongoDB 3.4 has been removed from Rocket.Chat 4.x](https://github.com/RocketChat/Rocket.Chat/pull/15199). We recommend using version 4.2.
 
+**Note** Rocket.Chat uses `mmapv1` storage engine which is **not** supported by Mongo versions >= 4.2.
```
debdutdeb, line 5: Rocket.Chat doesn't use any storage engine in specific. Rocket.Chat itself doesn't care about the mongodb storage backend. **Older** deployments do use the mmapv1 engine but one can easily migrate to wiredtiger and rc will work just fine.

blacksheep557: https://docs.rocket.chat/quick-start/installing-and-updating/manual-installation/ubuntu#configure-the-rocket.chat-service
but here it says that
`sudo sed -i "s/^#  engine:/  engine: mmapv1/"  /etc/mongod.conf`

debdutdeb: That section says that a replicaset is mandatory for rc > 1. Not mmapv1 :) But that article does need updating, thanks for pointing it out. I'll update it asap.

PR #2043, opened 2022-06-02
How to enable systemd template units

sdarwin:
Hi,

Reading the documentation, this text seemed to be logically inconsistent.  

```
The other Services will be enable since they are "WantedBy"=RocketChat.service
```
Imagine a user wants to enable services on ports 3002, 3003, and 3004.  Since these are systemd "templates", where will these numbers (3002, 3003, 3004) be stored and remembered later?  The answer, I think, is they will only be retained if you type a command such as 

```
systemctl enable rocketchat@3002
```

If that's the case, then "The services _may be_ enabled".   But not "the other services _will be_ enabled".

From https://unix.stackexchange.com/questions/506347/why-do-most-systemd-examples-contain-wantedby-multi-user-target

"However, WantedBy is separate from the enabled/disabled state: so in another sense, it's sort of a "preset": it determines under what conditions the automatic start may happen, but only when the service is enabled in the first place."

### DIFF
--- a/quick-start/installing-and-updating/other-deployment-methods/manual-installation/extras/multiple-instances-to-improve-performance.md
+++ b/quick-start/installing-and-updating/other-deployment-methods/manual-installation/extras/multiple-instances-to-improve-performance.md
@@ -67,7 +67,7 @@ WorkingDirectory=/path.to.rocketchat/rocket.chat
 ExecStart=/usr/local/bin/node /path.to.rocketchat/rocket.chat/bundle/main.js
 
 [Install]
-    WantedBy=rocketchat.service
+WantedBy=multi-user.target
 ```
 
 Start the other Rocket.Chat Services with
@@ -77,8 +77,8 @@ Start the other Rocket.Chat Services with
 If you want to run rocketchat at boot just enable the services with
 
 `systemctl enable rocketchat`
-
-The other Services will be enable since they are "WantedBy"=RocketChat.service
+or
+`systemctl enable rocketchat@3001`
 
 ### Ensure nodes can communicate
 


Rodriq (commented):
Remember we're considering the `rocketchat.service` like a main in this case. Any other service defined with the `WantedBy` value as the main service will automatically get started when the main service runs.

The link you shared, read from the second point
> if you omit the WantedBy=multi-user.target line and no other enabled service includes a Requires=your.service or Wants=your.service in its service definition, your service will not be started automatically.

A little about template specifiers https://www.digitalocean.com/community/tutorials/understanding-systemd-units-and-unit-files#template-specifiers
Please correct me if I'm wrong

sdarwin:
You wrote: "Any other service defined with the WantedBy value as the main service will automatically get started when the main service runs."     Not exactly, because the answer is more complicated than that.    

For a service to be automatically started, it needs to be **enabled**, and when it is **enabled**, that step looks to WantedBy for information.    Consider https://www.freedesktop.org/software/systemd/man/systemd.unit.html

WantedBy "A symbolic link is created in the .wants/ or .requires/ directory of each of the listed units when this unit is installed by **systemctl enable**."

So, WantedBy directs where the symbolic link is created.    After a symbolic link is created, that will cause a startup during boot. 
 Simply having WantedBy and nothing else, doesn't cause the automatic startup.

WantedBy + systemctl enable + symbolic link == automatic startup

Rodriq:
> You wrote: "Any other service defined with the WantedBy value as the main service will automatically get started when the main service runs." Not exactly, because the answer is more complicated than that.
> 
> For a service to be automatically started, it needs to be **enabled**, and when it is **enabled**, that step looks to WantedBy for information. Consider https://www.freedesktop.org/software/systemd/man/systemd.unit.html
> 
> WantedBy "A symbolic link is created in the .wants/ or .requires/ directory of each of the listed units when this unit is installed by **systemctl enable**."
> 
> So, WantedBy directs where the symbolic link is created. After a symbolic link is created, that will cause a startup during boot. Simply having WantedBy and nothing else, doesn't cause the automatic startup.
> 
> WantedBy + systemctl enable + symbolic link == automatic startup

I saw this note. In the link you shared
> "Note: WantedBy= and RequiredBy= are used in the [Install] section to create symlinks in .wants/ and .requires/ directories. They cannot be used directly as a unit configuration setting.

 Clear now! Thanks
 The docs will be updated

PR #2046, opened 2022-06-07
client_max_body_size setting in nginx

sdarwin:
Hi,

From https://github.com/RocketChat/Rocket.Chat/issues/13351 it was necessary to set `client_max_body_size 0;` to fix the bug 
"Failed_To_upload_Import_File" for Slack Import.      I encountered the issue and it's also corroborated by other users in the discussion.   

"By default, NGINX® has a upload limit of 1 MB per file. By editing client_max_body_size, you adjust the file upload size."


Rodriq (approved):
Setting `client_max_body_size` to `0`, the client request body size is not checked therefore making it unlimited.
Good pointing out 😄  Docs updated.

PR #2061, opened 2022-07-11
Update auto-ssl-with-snaps.md

mholt:
Hi!

I'm looking at using Rocket.chat for a family thing, and remembered that Rocket.chat uses Caddy. So I looked at the docs and realized some of it was out of date (for about 2 years...)

I updated Caddy config to Caddy 2 format, and made consistent usage of TLS instead of "SSL". Corrected some slight nuances about auto TLS.


Rodriq:
Thanks for the update. @mholt

PR #2088, opened 2022-12-14
Rollbar Rocketchat integration

TheDevFactory:


barbarazanella:
Hello, TheDevFactory!

We love your idea and want to suggest that you turn it into an app that can be available on our marketplace. With this initiative, we can reduce the complexity of the core and provide a customization experience for Rocket.Chat users.

What do you think about that?

Let us know if you need any help; you can be part of our publisher's community :D

